PR #5, opened 2013-10-07
C# TLS impementation updated by fresh port from the BC Java. DTLS and Sever side support introduced.

ArmanNX:


peterdettman:
We didn't apply this patch, but AFAICT all the relevant functionality was ported. If there's anything still missing it'll have to be addressed in a new pull request.

PR #15, opened 2014-06-08
Update for use of higher Hashing Algorithms for key Hash

darkoperator:
Currently it supports only the use of SHA1. The latest changes in RFC 4880 in section 13.6 for DSA http://tools.ietf.org/html/rfc4880#section-13.6 if a value for DSA higher than 1024 us used SHA1 is no longer valid this will also affect RFC 6637 where it would be SHA 256. This changes should allow for use of a higher Hashing algorithm for hashing.


peterdettman:
Merged, thanks. I took out the Dispose method as we don't use them anywhere in the code at present. (This will change after 1.8, as we address PCL support).

PR #16, opened 2014-06-22
Permit Org.BouncyCastle.Crypto.Generators.ECKeyPairGenerator to save cur...

kyanha:
...ve names, per Jira BMA-101

When publicKeyParamSet is left null, then GenerateKeyPair() does not look up the parameter set to try to save the curve name.  When it's set, the constructor overloads for ECPublicKeyParameters and ECPrivateKeyParameters take the curve name and don't bother with explicit encoding.

This resolves Jira issue BMA-101.

### DIFF
--- a/crypto/src/asn1/cms/EnvelopedData.cs
+++ b/crypto/src/asn1/cms/EnvelopedData.cs
@@ -94,7 +94,7 @@ namespace Org.BouncyCastle.Asn1.Cms
                 return (EnvelopedData)obj;
             if (obj == null)
                 return null;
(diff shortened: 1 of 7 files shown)


peterdettman:
I've finally applied the core (one-liner) fix, thanks. The BigInteger changes won't work in .NET 1.1; it will probably be feasible after the next release when we are intending to move to a much more recent .NET version. The "GetInstance" changes actually introduce recursions... the warnings will go away when those constructors are changed to private (they are marked Obsolete because they shouldn't be public), again probably after the next release.

PR #22, opened 2014-11-23
Updated project files to build using Visual Studio 2010+

jstedfast:
Target .NET 2.0 instead of 1.1 since no one uses 1.1 anymore.

Also don't compile the unit tests & data into the crypto assembly, instead create a separate UnitTests project/assembly for those.

This makes it easier to integrate BouncyCastle into other projects and builds nicely with both Xamarin Studio (aka MonoDevelop) and Visual Studio 2010+.


jstedfast:
Ping.

peterdettman:
@jstedfast I had thought maybe just to jump straight to the new VS 2013 community edition, thoughts? I am interested in compatibility with Xamarin too, if that is a factor.

jstedfast:
I think that targeting the VS 2013 Community Edition would probably be a good idea (and the .csproj files in the patches above will work with VS 2013 Community Edition).

When I say that they are in the VS 2010 format, it's the same format that is used for VS2010, VS2012, and VS2013 (and I think the upcoming VS2015 as well, but I'm not 100% sure about that).

timw:
This is blocking new devs a bit now - MonoDevelop/Xamarin Studio has dropped support for the .mds projects entirely (not even an import function).

jstedfast:
MonoDevelop actually dropped support for mdw/mds 4 years ago :)

peterdettman:
@jstedfast We'll be putting out 1.8 RC2 either tonight or tomorrow. There'll be a week or so then before the final release (or longer if there are bugs of course). If this PR is brought up to date with RC2, I will be happy to merge it prior to full release.

jstedfast:
Awesome, thanks Peter. I'll update the branch this weekend.

peterdettman:
Merged, with a few updates. Thanks for your patience.

jstedfast:
Awesome, thanks Peter!

PR #27, opened 2015-02-23
Update EncryptedData.cs

nikodz:
This update is about unprotected attributes.

According to https://tools.ietf.org/html/rfc4134#page-114 unprotected attribute is sequence in tagged object, not set. Also, parser was looking for set instead of tagged object.

### DIFF
--- a/crypto/src/asn1/cms/EncryptedData.cs
+++ b/crypto/src/asn1/cms/EncryptedData.cs
@@ -7,7 +7,7 @@ namespace Org.BouncyCastle.Asn1.Cms
 	{
 		private readonly DerInteger				version;
 		private readonly EncryptedContentInfo	encryptedContentInfo;
-		private readonly Asn1Set				unprotectedAttrs;
+		private readonly Asn1Sequence				unprotectedAttrs;
 
 		public static EncryptedData GetInstance(
 			object obj)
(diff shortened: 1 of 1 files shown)


bcgit:
The issue is that the object is implicitly tagged so in a dump will often be interpreted as a sequence as the actual type is lost and the only thing that can be recognised is that it is a constructed type - at that level calling it a sequence is as good as any.

For the actual definition see RFC 5652 section 6.1 which is based on the original PKCS#7 definition, which is:

      EnvelopedData ::= SEQUENCE {
        version CMSVersion,
        originatorInfo [0] IMPLICIT OriginatorInfo OPTIONAL,
        recipientInfos RecipientInfos,
        encryptedContentInfo EncryptedContentInfo,
        unprotectedAttrs [1] IMPLICIT UnprotectedAttributes OPTIONAL }

UnprotectedAttributes ::= SET SIZE (1..MAX) OF Attribute

Having said that the call to getInstance() should be modified to recogninse the implicit tagging.

nikodz:
Definition you referenced "RFC 5652 section 6.1" is EnvelopedData and I am talking about EncryptedData.
You are right, it's implicitly tagged and type is not defined, but inside of that tag there are sequences where every sequence is unprotected attribute.

bcgit:
Yes, sorry, should have used EncryptedData, it's refering to the same structure though.

peterdettman:
@nikodz We did realize the confusion yesterday and in fact have patched both EncryptedData.cs and EncryptedData.java in the latest git code (to match the way it's parsed in EnvelopedData). I think it was missed originally because EncryptedData is not really used much, but if you do in fact need to use it, prior to the forthcoming 1.8.0 C# release, I guess you can either take the latest code, or just make a local copy of the patched EncryptedData.cs .

nikodz:
@bcgit @peterdettman Thanks. I am actually using encrypted data on C#. I just took latest code and it works perfectly.

peterdettman:
@nikodz Thanks for the confirmation (and for reporting the problem).

PR #29, opened 2015-03-24
Correct duplicate key exception when more than one signer use the same digest algorithm.

mps-beppler:
When sigin a document, if two or more signers use the same algorithm, an duplicate key exception is generated.

This commit solves the problem.


peterdettman:
Looks good, thanks for the patch. Merged.

PR #32, opened 2015-04-10
Support (low-level) for non-default PRFs when using PKCS5v2 

bartland:
This is a minimal change where asn1/pkcs/PBKDF2Params.cs is modified in a similar manner as in bc-java.

This allows low level routines to be used to support SHA256, etc. in particular, using Pkcs5S2ParameterGenerator().

Use via the high level routines (security/PbeUtilities.cs) is not currently supported.

These changes have been cross-tested with bc-java and with openssl 1.0.1j.


peterdettman:
Merged with minor changes, thanks!

Issue #34, opened 2015-06-16
Unresolved critical extensions issue in PkixCertPathChecker 


marcinpaszylk:
When overriding Check method of PkixCertPathChecker.cs I can't remove unresolved critical extensions I don't want to check. 
Reson: unresolvedCritExts object is immutable ICollection.
Fix: It could be of type Org.BouncyCastle.Utilities.Collections.ISet, this way I can call Remove() method on collection.

peterdettman:
Addressed by #35.

PR #35, opened 2015-06-16
Changed PkixCertPathChecker Check method signature, so chosen unsuppo…

marcinpaszylk:
…rted certificate extensions can be removed from collection in Check method implementation.


marcinpaszylk:
fixes #34

peterdettman:
Merged. Thanks for the patch.

marcinpaszylk:
Hi,
when could I expect new version of the library with this fix?

peterdettman:
In the next few weeks, sorry I can't be more specific.

PR #40, opened 2015-10-19
Bug of string comparison for hungarian culture

NicolasDorier:
Detailed here : http://stackoverflow.com/a/29933353

> The mechanism string is "HMACSHA256" and the if statement never evaluates true for me, because my language is Hungarian and "CS" is a letter in Hungarian. So according to the StartsWith function "HMACSHA256" doesn't start with "HMAC" in Hungarian.

### DIFF
--- a/crypto/src/security/MacUtilities.cs
+++ b/crypto/src/security/MacUtilities.cs
@@ -112,15 +112,15 @@ namespace Org.BouncyCastle.Security
                 mechanism = upper;
             }
 
-            if (mechanism.StartsWith("PBEWITH"))
+            if (mechanism.StartsWith("PBEWITH", StringComparison.Ordinal))
             {
                 mechanism = mechanism.Substring("PBEWITH".Length);
             }
 
-            if (mechanism.StartsWith("HMAC"))
+            if (mechanism.StartsWith("HMAC", StringComparison.Ordinal))
             {
                 string digestName;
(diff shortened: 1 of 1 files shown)


NicolasDorier:
Here simple repro
```
Thread.CurrentThread.CurrentCulture = new CultureInfo("hu"); 
string mechanism = "HMACSHA256"; 
var v1 = mechanism.StartsWith("HMAC");

mechanism = "HMAC-SHA256"; 
var v2 = mechanism.StartsWith("HMAC");
```

jstedfast:
You should probably just use `StringComparison.Ordinal` instead of `OrdinalIgnoreCase`, but otherwise, this looks like the correct type of fix to me (although I did notice some accidental re-indents of some lines of code).

NicolasDorier:
Fixed and squashed.

NicolasDorier:
@peterdettman can you merge ? this is a serious bug.

peterdettman:
It will be fixed for 1.8, but it potentially affects more than just this file, so we are reviewing.

peterdettman:
The proposed overload of String.StartsWith is not available in .NET 1.1, so I've used:
           CultureInfo.InvariantCulture.CompareInfo.IsPrefix(mechanism, "HMAC", CompareOptions.Ordinal)

A fix for the immediate problem has been committed. We will follow up with a more general refactoring of the various other *Utilities classes and a review of other String methods in the codebase.

peterdettman:
Finished the followup changes for various String methods that default to culture-sensitive.

PR #49, opened 2016-01-13
Implemented passing public key bitstring when generating EC private key info

KimikoMuffin:
Compatibility with OpenSSL, SSH, PuTTY, and pretty much everything else that exports ECDSA keys as PEM files. Previously, BouncyCastle omitted the encoding of the public key, which meant that they couldn't be loaded anywhere else.


peterdettman:
This change has finally been applied, thanks.

Issue #53, opened 2016-02-05
Add support for mixed hashing (RSA/ECB/OAEPWithSHA-256AndMGF1Padding)


wizche:
In java (up to JDK8) per default the MGF digesting algorithm used is SHA1, currently there is no way to specify a different hashing algorithm between OAEP and MGF.

One (bad way) to circumvent this is to use reflection to change mgfHash field in the cipher:

``` c#
var cipherImpl = GetInstanceField(c.GetType(), c, "cipher");
var newDigest = new Sha1Digest();
var prop = cipherImpl.GetType().GetField("mgf1Hash", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance);
prop.SetValue(cipherImpl, newDigest);
```

fabiogermann:
maybe this pull request will help: https://github.com/bcgit/bc-csharp/pull/54/files

bcgit:
Just one thing with this one - which Java provider is being talked about here? The BC one, in line with the recommended practice with OAEP, defaults to the same mask generation digest as the base one.

Fabio's patch will obviously help with this, but I am wondering about the usage at the Java end - if possible OAEP should be configured to use the same digest in both roles.

wizche:
I'm talking about the SUN provider, which, when using the default cipher initialisation, doesn't use the same digest algorithm for both OAEP and MGF. You are completely right that changing the configuration on the java side (setting both digest the same) is the best solution. But as you know, sometimes legacy application need much time to perform such a change...

I tested the pull request of Fabio and worked as intended. Maybe you could add a warning when using incongruent digests that suggest the users to be compliant with recommendation at some point.

wizche:
BTW: There is an open issue in OpenJDK targeting this very problem: https://bugs.openjdk.java.net/browse/JDK-7038158

fabiogermann:
Is there anything missing in this issue and related pull request? I know it's an old one but still would be great to have this merged.

dghgit:
So I've just checked and the sun provider is still doing this. I'd say this issue will never be fixed, the other weird thing the cipher implementation is using BER for the parameters encoding... Anyway, I've merged the patch.

PR #54, opened 2016-02-18
added an implementation of RSA/ECB/OAEPWithSHA-256AndMGF1Padding as …

fabiogermann:
… it is implemented in java as: RSA/ECB/OAEPWithSHA-256AndMGF1WithSHA-1Padding

**Added compatibility with Java:**

In Java this Cipher : `RSA/ECB/OAEPWithSHA-256AndMGF1` has the MGF1 Padding "SHA1". In C#
the same Cipher `RSA/ECB/OAEPWithSHA-256AndMGF1` has the MGF1 Padding "SHA256".

The current BC C# implementation is unable to de/en-crypt data in such a way that it would be compatible with a Java counterpart.

Therefore the `RSA/ECB/OAEPWithSHA-256AndMGF1WithSHA-1Padding` configuration has been added to allow this interoperability. For completeness the `RSA/ECB/OAEPWithSHA-256AndMGF1WithSHA-256Padding` has been added as well.

Related to https://github.com/bcgit/bc-csharp/issues/53

### DIFF
--- a/crypto/src/security/CipherUtilities.cs
+++ b/crypto/src/security/CipherUtilities.cs
@@ -85,6 +85,10 @@ namespace Org.BouncyCastle.Security
             OAEPWITHSHA_224ANDMGF1PADDING,
             OAEPWITHSHA256ANDMGF1PADDING,
             OAEPWITHSHA_256ANDMGF1PADDING,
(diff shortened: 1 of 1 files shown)


echoorchid:
wow, good !!!

dghgit:
Thanks for the patch, now merged.

PR #63, opened 2016-06-13
CmsSignedData: Don't assume the signed content is a Asn1OctetString object

qmfrederik:
This PR fixes an issue in the `CmsSignedData` where the constructor would assume that the signed content is an `Asn1OctetString` object. That's not always true - in my case, the data was a `DerSequence`.

### DIFF
--- a/crypto/src/cms/CMSSignedData.cs
+++ b/crypto/src/cms/CMSSignedData.cs
@@ -135,13 +135,10 @@ namespace Org.BouncyCastle.Cms
 			//
 			if (signedData.EncapContentInfo.Content != null)
 			{
+				var content = signedData.EncapContentInfo.Content;
 				this.signedContent = new CmsProcessableByteArray(
-					((Asn1OctetString)(signedData.EncapContentInfo.Content)).GetOctets());
+					content.GetEncoded());
 			}
-//			else
-//			{
-//				this.signedContent = null;
-//			}
 		}
 
 		/// <summary>Return the version number for this object.</summary>


peterdettman:
The proposed change would change (break) the behaviour when the signed content is an Asn1OctetString, since GetOctets is not the same as GetEncoded.

The BC Java version has extra code to allow a sequence to be handled here, which would be reasonably straightforward to port over, but it is not quite as simple as calling GetEncoded on the sequence (in fact it appears to just encode the elements of the sequence). So I doubt if the proposed change is correct for sequences either.

It would be helpful if you could give more information on what type of files you are trying to work with, and if you know it, the relevant applicable RFC or other standards document that we could refer to.

nmoinvaz:
I think [RFC 5652 section 5.2.1](https://tools.ietf.org/html/rfc5652#section-5.2.1), Compatibility with PKCS #7, applies here as well:

> PKCS #7 defines content as:

> content [0] EXPLICIT ANY DEFINED BY contentType OPTIONAL

> The CMS defines eContent as:

>  eContent [0] EXPLICIT OCTET STRING OPTIONAL

Genbox:
Authenticode signed files are PKCS#7 and not CMS, which will result in BC breaking inside CmsSignedDataParser. BC Java Beta (156b03) have implemented PKCS#7 support, which should solve this.

peterdettman:
I think this was addressed by https://github.com/bcgit/bc-csharp/commit/2a508f3ffe7843efcd188f6349f3125a25158351 .

PR #64, opened 2016-06-13
Set signature through ISignatureFactory

alexander-irrgang:
Using Generate with ISignatureFactory is the preferred way to use the generator. However the signature could only be set using obsoleted methods.
The change just uses the same signature algorithm information provided by the ISignatureFactory, which was already used in line 179 beforehand.

### DIFF
--- a/crypto/src/x509/X509V2AttributeCertificateGenerator.cs
+++ b/crypto/src/x509/X509V2AttributeCertificateGenerator.cs
@@ -164,6 +164,7 @@ namespace Org.BouncyCastle.X509
 				acInfoGen.SetExtensions(extGenerator.Generate());
 			}
 
+          acInfoGen.SetSignature((AlgorithmIdentifier)signatureCalculatorFactory.AlgorithmDetails);
 			AttributeCertificateInfo acInfo = acInfoGen.GenerateAttributeCertificateInfo();
 
             byte[] encoded = acInfo.GetDerEncoded();


DmitriNymi:
@alexander-irrgang , do you have any experience with signing Pkcs10CertificationRequestDelaySigned with ECDSA using CSR (certificate signing request) hash ?

I posted a question: https://stackoverflow.com/questions/46182659/how-to-delay-sign-the-certificate-request-using-bouncy-castle-with-ecdsa-signatu

It worked for RSA but not ECDSA:

        /// hash Algorithm Oid, for example: "2.16.840.1.101.3.4.2.1"
        /// signAlgorithm = "RSA" or "ECDSA"
        /// <param name="privateSigningKey">private key for signing</param>
        public static byte[] Sign(byte[] hash, string signerAlgorithm, string hashAlgorithmOid, AsymmetricKeyParameter privateSigningKey)
        {
            
            var digestAlgorithm = new AlgorithmIdentifier(new DerObjectIdentifier(hashAlgorithmOid), DerNull.Instance);
            var dInfo = new DigestInfo(digestAlgorithm, hash);
            byte[] digest = dInfo.GetDerEncoded();

            ISigner signer = SignerUtilities.GetSigner(signerAlgorithm);
            signer.Init(true, privateSigningKey);
            signer.BlockUpdate(digest, 0, digest.Length);
            byte[] signature = signer.GenerateSignature();
            return signature;

peterdettman:
This fix has finally been applied, thanks.

PR #65, opened 2016-06-17
SignerInfo: Only write attribute object if it contains data

qmfrederik:
If a `CmsSignedObject` is signed without any attributes (as seen, for example, for `.apk` packages), Bouncy Castle would write out a 0-length object. That doesn't seem to be valid; you should omit the object all together.

This PR addresses that issue.

### DIFF
--- a/crypto/src/asn1/cms/SignerInfo.cs
+++ b/crypto/src/asn1/cms/SignerInfo.cs
@@ -167,14 +167,14 @@ namespace Org.BouncyCastle.Asn1.Cms
             Asn1EncodableVector v = new Asn1EncodableVector(
                 version, sid, digAlgorithm);
 
-            if (authenticatedAttributes != null)
+            if (authenticatedAttributes != null && authenticatedAttributes.Count > 0)
             {
                 v.Add(new DerTaggedObject(false, 0, authenticatedAttributes));
             }
 
             v.Add(digEncryptionAlgorithm, encryptedDigest);
 
-            if (unauthenticatedAttributes != null)
+            if (unauthenticatedAttributes != null && unauthenticatedAttributes.Count > 0)
             {
                 v.Add(new DerTaggedObject(false, 1, unauthenticatedAttributes));
             }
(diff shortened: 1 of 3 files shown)


peterdettman:
Thanks for the PR. It does appear from RFC 5652 that the attribute sets are to have a minimum size of 1, and one infers they should be absent if they would otherwise be empty. However, RFC 2315 (PKCS#7) isn't as clear on the matter, and AFAIK people still use the BC APIs for PKCS#7. So we might prefer to leave the control there for the user; I'll discuss it with the team.

In case you need a workaround, are you aware that there are overloads for CmsSignedDataGenerator.AddSigner that allow to pass a custom CmsAttributeTableGenerator for the signed/unsigned attributes? I believe if null is passed for one of those, there won't be any actual SET written to the ASN.1.

I also think it would be an improvement to allow a null return from CmsAttributeTableGenerator.GetAttributes, which would allow the same behaviour as a dynamic decision.

qmfrederik:
Based on the tests I've done, if you pass `null` as a `CmsAttributeTableGenerator` to the `CmsSignedDataGenerator.AddSigner` method, you still end up with some attributes which are generated by default.

I had to pass my own `CmsAttributeTableGenerator` like this and pass that one to `AddSigner` for it to work:

```
var unsignedAttributeTableGenerator = new SimpleAttributeTableGenerator(new AttributeTable(new Asn1EncodableVector()));
```

jimsch:
For the authenticated attribute the field should be omitted and this is very clear in RFC 2315.

authenticatedAttributes is a set of attributes
             that are signed (i.e., authenticated) by the signer. The
             field is optional, but it must be present if the content
             type of the ContentInfo value being signed is not data. If
             the field is present, it must contain, at a minimum, two
             attributes

The last sentence says that it cannot be an empty sequence.
It is less clear from the document if unauthenticated attributes should be absent if empty, but every package I have ever used did that.  Also it would be along the same lines as what is done for authenticated attributes.

bcgit:
qmfrederick - I think in this case you are talking about a direct signature, if you use SignerInfoGeneratorBuilder.SetDirectSignature(true) you will get the result you are looking for. Note that the calculation of the signature in this case is done on a digest of the actual data.

The old AddSigner methods will be marked obsolete so I would recommend this approach in general.

dghgit:
Closed as appears resolved by use of direct signature.

PR #66, opened 2016-06-18
Enable subclassing of the DerOutputStream

qmfrederik:
I use Bouncy Castle to generate signatures for iOS and Android applications (as PKCS#7 objects).
It turns out that these mobile operating systems are very sensitive to how these signatures are serialized to disk.
For example, Android seems to bark at any object with undefined length, whereas iOS can't handle objects with a length greater than 4096 _unless_ they are marked as undefined length.

This isn't something that can be fixed in Bouncy Castle itself, as all of these are implementation details which will differ between the systems you interact with.

Hence, I propose to enable subclassing of the `DerOutputStream`, so that those who want greater control of how PCKS#7 objects (or any other object, for that matter) are serialized to disk can subclass  `DerOutputStream` and implement the required format themselves.

This PR implements that.

Makes sense?

### DIFF
--- a/crypto/src/asn1/DerOutputStream.cs
+++ b/crypto/src/asn1/DerOutputStream.cs
@@ -39,7 +39,7 @@ namespace Org.BouncyCastle.Asn1
 			}
 		}
 
-		internal void WriteEncoded(
+		public virtual void WriteEncoded(
 			int		tag,
 			byte[]	bytes)
 		{
@@ -48,7 +48,7 @@ namespace Org.BouncyCastle.Asn1
 			Write(bytes, 0, bytes.Length);
 		}
 
-        internal void WriteEncoded(
+        public virtual void WriteEncoded(
             int     tag,
             byte    first,
             byte[]  bytes)
@@ -59,7 +59,7 @@ namespace Org.BouncyCastle.Asn1
             Write(bytes, 0, bytes.Length);
         }
 
-        internal void WriteEncoded(
+        public virtual void WriteEncoded(
 			int		tag,
 			byte[]	bytes,
 			int		offset,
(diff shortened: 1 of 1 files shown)


qmfrederik:
@peterdettman Any chance we could get this merged? It looks like a simple change and would prevent me from having to roll my own version of BouncyCastle.

Happy to help in any way I can.

peterdettman:
After reviewing the ASN.1 code and how the DerOutputStream class is used, I don't think this is a great idea.

- Various of the BER type classes (BerSequence, BerSet) check the concrete type of the DerOutputStream (using 'is' tests) to determine whether they should force DER encoding.
- These DerOutputStream methods were only extracted as common code from the BER/DER type classes, and they are not consistently used, so you can't override e.g. WriteNull and expect that you are intercepting all the places where ASN.1 NULL is encoded.
- Turning the previous point around, most encodings are generated by the BER/DER type classes themselves and they call the DerOutputStream utility methods to do very specific things, not as an abstraction. If the methods weren't guaranteed to do exactly that, they wouldn't be calling them.

So I suspect you are overestimating the usefulness of being able to override these methods, possibly imagining that DerOutputStream is a central place to be able to control how ASN.1 objects are encoded.

We are well aware that the ASN.1 code has its limitations. The current C# version is what could be called "first generation". The Java version has moved to a somewhat better API, but it's still by no means perfect, and such a transition is not without pain for users.

If the Android/iOS example is the specific problem you are trying to solve, please open a new issue for it and give more details, we can probably address it at a different level.

dghgit:
Closing - will review #88 but a quick look at #88 makes it look like the issue was resolved.

Issue #50, opened 2016-01-16
Thoughts on code modernization?


jarz:
Some idioms seem outdated (implementing IEnumberable but not IEnumerable<T> being a big one).

Is this just an effect of making the C# and Java code look similar? Would clean-up contributions be accepted?

Thanks!

peterdettman:
It's partly the Java origins, partly backward compatibility, partly supporting .NET 1.1 (1.8 series is the last release for that). We will be forking off a maintenance branch for 1.8 "shortly", at which point we would welcome modernization patches to the master branch (intended target .NET 4).

onovotny:
@peterdettman is there any point in using .NET 4 as the baseline given that .NET 4-4.5.1 is no longer supported by Microsoft for security updates?

Can the baseline be .NET 4.5.2 (or at least 4.5 from an API/compat standpoint)?

jstedfast:
FWIW, I'm fine with Bouncy Castle deciding to target >= .NET 4.5, but here are a few thoughts:

1. There's no sense in targeting a .NET platform version higher than what you need API-wise.
2. .NET 4.0 gives you everything you need to create Task-based API's allowing developers consuming your library to take advantage of async if they want to (granted, if BC itself wants to take advantage of doing things asynchronously internally, that would be grounds for depending on .NET 4.5 imho)

So I guess the question comes down to: what does .NET 4.5 offer that we'd want to use inside BC?

.NET 2.0 offers a LOT of advantages over 1.1 such as the Dispose pattern for Streams, generics, etc.
.NET 3.5 starts to add some LINQ which can be really handy
.NET 4.0 adds some nice Stream API's such as CopyTo(Stream) (or was this added in .NET 3.5?), CancellationToken, System.Threading.Task, etc.
.NET 4.5 adds async/await, but unless bouncy castle actually intends to use async internally, this is somewhat moot.

onovotny:
.NET 4.5's most important characteristic is that it's a `System.Runtime` based target. That is, it can use contracts instead of mscorlib directly.

More to the point, you'll be able to use xProj to create multiple profiles in a single project:
  - `netstandard1.3` full support for everything and consumable by .NET Core
  - `net45` support for .net 4.5+
  - `portable-win8+wpa81+wp8+net45` Profile 259 with some features not present (filesystem, etc)

Then you just have iOS/Android projects until they support `netstandard1.3`. Once Xamarin supports those contracts you won't need platform specific builds for them.

[...]

PR #68, opened 2016-12-03
Updates to support .NET Standard and multi-targeting

### LINKED ISSUE #41: Move to netstandard TFM
Moving this issue here from https://github.com/onovotny/BouncyCastle-PCL/issues/12

This is to move from `dotnet` to `netstandard` in the nuspec. See https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/standard-platform.md for details.

The nuspec will be greatly simplified as it required extra blank dependency groups and copies of the libs in platform-specific TFM `\lib` dirs. to work correctly.

This issue is dependent on https://github.com/dotnet/corefx/issues/4367 being completed.

/cc @peterdettman

clairernovotny:
This is the latest branch of the PCL version, updated to support:
- `net6.0`
 - `net4`
-  `netstandard2.0`

That provides the widest range of functionality to supported platforms, including Xamarin and .NET Core. The `netstandard2.0` version has the exact same surface area as the desktop `net4` version.

Fixes #41


jstedfast:
👍

jstedfast:
@onovotny now that you are targeting .NETStandard 1.x, it should be possible to re-enable the DotNetUtilities class, right? Specifically, it'd be awesome to re-expose the methods that convert back and forth between BC and System.Security X509Certificates.

If I send you a PR, where should that go?

jstedfast:
Looks like I actually only need 2 methods: `GetKeyPair` and `FromX509Certificate`.

onovotny:
a PR should go to the netstandard branch of my fork

jstedfast:
Looks like it's not possible even in netstandard1.3. I think I may just have to copy the `DotNetUtilities` class into MimeKit for use with Xamarin-specific targets if I want to expose the API's that allow users to pass in System.Security.Cryptography.X509Certificates.

The `RSA` and `DSA` classes needed for `GetKeyPair` don't exist in netstandard1.3.

No worries. I was hoping I wouldn't need to do that, but alas...

onovotny:
@jstedfast what netstandard version are they in? I personally have no objections to including a higher version if there's a need. Multi-targeting is easy :)

jstedfast:
I'm actually also targeting netstandard1.3 for MimeKit and MailKit. I just also have Xamarin-specific targets.

I just realized, however, that Mono's X509Certificate2 implementation doesn't include the PrivateKey anyway, so `GetKeyPair` is useless on the Xamarin platforms and the 1 case where all I need is the certificate, `DotNetUtilities.FromX509Certificate()` is just 1 line of code, so doing that "manually" inside MimeKit is not an issue.

jstedfast:
I appreciate the offer, though!

onovotny:
@jstedfast it's not hard to also target netstandard1.5/1.6 if that helps? better than Xamarin specific targets.

martinstoeckli:
Did you ever think about creating a [shared project](https://developer.xamarin.com/guides/cross-platform/application_fundamentals/shared_projects/)? Shared projects are supported since Visual Studio 2013 Update 2 and as long as they do not have dependencies, they are a godsend for cross platform development.
No problems with dotnet versions/targets anymore, no problems with assembly versions or with loading libraries, just compiling and be done. Btw thanks for your efforts, they are really appreciated.

onovotny:
@martinstoeckli I am familiar with shared projects, but what benefits would you see in this case over simply multi-targeting? One benefit I see with multi-targeting and the SDK-style projects are that all of the files can stay as-is, making it much easier to maintain.

martinstoeckli:
There are some many variations of targets (dotnet version, OS platform, 32-64 bit, .Net core, ...) that I often get into the situation, that external libraries are mutually exclusive and I cannot use them. A shared project, as long as it doesn't have any problematic dependencies, is compiled to the same target as the app itself and therefore is always compatible. The multi-target libraries have to be maintained for every future target, a shared library is pure code. Of course this is a bit simplified and I must admit that I didn't have any problems with bc-csharp library so far. Thanks again.

penartur:
@martinstoeckli ,
> There are so many variations of targets (dotnet version, OS platform, 32-64 bit, .Net core, ...) that I often get into the situation, that external libraries are mutually exclusive and I cannot use them.

Targeting .NET Standard 1.0 immediately gives support of .NET Core, .NET Framework 4.5+, Mono 4.6+ and a lot of other platforms. Building for "AnyCPU" will make library compatible with x86, x64, ARM etc.

Library multi-targeting .NET Standard 1.0 and .NET Framework 1.1 won't be mutually exclusive with anything else, until you have some unusual use case (e.g. your application needs to work with Xamarin.Mac 2.0).

onovotny:
@jstedfast circling back on this, is there anything specific you want/need in a netstandard2.0 version?

jstedfast:
@onovotny awesome! (sorry, just saw your comment now)

TalAloni:
Thanks for your efforts, can you please not delete BouncyCastle.csproj? Some of us still need .Net Framework 2.0 support!

onovotny:
All target frameworks can be reached from the current project format.

What project is only targeting 2.0 *and* is using the latest libraries? Projects that old are typically locked down and don't randomly update dependencies.

TalAloni:
1. Are you saying that I can still build a .Net Framework 2.0 compatible assembly with your proposed project format? could you please verify that?
2. Why there are two 'crypto.csproj' in your fork? (crypto/crypto.csproj and src/crypto/crypto.csproj)
3. There are many reasons to target the .Net Framework 2.0 on new projects, I'm not saying we should always cling to old versions at all cost, and I'm all for dropping .Net 1.1 support and using generics to improve the library and APIs, but if we could have .Net Framework 2.0 support at the cost of an additional project file it seems well worth it to me.

onovotny:
Yes, just add net20 to the TargetFrameworks list. 

NET 2.0 is 12 years old. I really don't see why current libraries need to continue targeting it.

What reason is there to target 2.0 on a new project, I really can't think of any.

The other csproj is there to minimize merge conflicts for now.

jstedfast:
Microsoft doesn't support anything older than, I think, .NET 4.6.1 at this point. Realistically, it probably doesn't make any sense to target anything older than 4.5. Every modern Windows system dating back to Vista SP2 supports .NET 4.5 so there's no reason not to.

onovotny:
Technically, net 4.5 and 4.5.1 aren't supported either. 4.5.2 is the lowest 4.x version getting patches.

3.5 SP1 is still getting patches too as it's a Windows component. I still would not use it for any new project :)

TalAloni:
1. If you need an additional 'crypto.csproj' then please name it appropriately and put it where it belong (not in the 'src' directory)

2. While reviewing some of the 50 commits you pulled, I have noticed many changes to indentation (tabs become spaces or vice versa), while currently (and unfortunately) there is no consistency in the BC source code, IMO these changes should not be part of the current PR.
also, I'm assuming you're going to rebase those 50 commits eventually.

3. We seem to be having the wrong discussion here, support for .NET Standard and support for .Net Framework 2.0 / 3.0 / 3.5 are not mutually exclusive.

4. I tried to open the new project file but it seems that VS2017 is required, can you keep a legacy project file to prevent us from the need to upgrade? (which for some of us has costs involved)

Again, I really appreciate the amount of work you've done here, and I'm all for .NET Standard, let's try to figure out a way to prevent this from being a breaking change, let me know if I can help in any way.

onovotny:
> If you need an additional 'crypto.csproj' then please name it appropriately and put it where it belong (not in the 'src' directory)

This is following the recommended .NET project system. src contains the actual source files and test contains the test files. Each should have a project in them as the project files have default globs that include all files underneath. Having one above src means that src and test are in the same project, which is not the case.

> While reviewing some of the 50 commits you pulled, I have noticed many changes to indentation (tabs become spaces or vice versa), while currently (and unfortunately) there is no consistency in the BC source code, IMO these changes should not be part of the current PR.
also, I'm assuming you're going to rebase those 50 commits eventually.

I was hoping to avoid rebasing as this branch has lived too long. If a final rebase is necessary, then I'll do it just before it's merged. That still won't change the tabs/spaces issues. The only real way to address that is with an `.editorconfig` setting. BTW, VS 2017 includes support for that built-in.

> We seem to be having the wrong discussion here, support for .NET Standard and support for .Net Framework 2.0 / 3.0 / 3.5 are not mutually exclusive.

Fair. I personally don't see the need in continuing it, but it's not hard to add an additional target for it.

> I tried to open the new project file but it seems that VS2017 is required, can you keep a legacy project file to prevent us from the need to upgrade? (which for some of us has costs involved)

VS 2017 is fully required due to the significant enhancements in the build system. It is not possible to keep an older version. VS 2017 Community edition is free for OSS use and can be installed alongside any other VS edition.

Furthermore, if the legacy projects are dropped (.NET 4, Portable-) and .NET 4.6 being the minimum, then `dotnet` can build it on any platform. Currently, `MSBuild.exe` is required due to the older platforms being targeting.

yaakov-h:
Is there any timeline for merging this?

SeriousM:
Any updates on the release of 1.9.0 which will target .netstandard?

SeriousM:
@peterdettman, @onovotny: is there any eta when this package is available as (beta) nuget? Thanks

SeriousM:
> It's already available here: https://www.nuget.org/packages/Portable.BouncyCastle/

Thank you. I was expecting to see the .netstandard support on the official package https://www.nuget.org/packages/BouncyCastle/ which states: 

> This is the final feature release with support for legacy .NET platforms. From 1.9.0 we will be targeting more modern .NET frameworks (see https://github.com/bcgit/bc-csharp/pull/68) and updating our build and packaging systems. The 1.8.x series will continue to receive bug fixes, but limited new functionality.

Therefor I was on the lookout for the version 1.9.0

Thank you for clarification. I'm quite disappointed about the communication from the official bc-team :(

Pzixel:
Any progress on that?

jariq:
@Pzixel following statement can be found in [Release Notes for 1.8.5](https://www.bouncycastle.org/csharp/index.html#RELEASENOTES185) available at bouncycastle.org:

> This is the final feature release with support for legacy .NET platforms. From 1.9.0 we will be targeting more modern .NET frameworks (see https://github.com/bcgit/bc-csharp/pull/68) and updating our build and packaging systems. The 1.8.x series will continue to receive bug fixes, but limited new functionality.

Pzixel:
I read this. I'd like to get what are plans for 1.9.X release date? I have to perform some keccak encryption before mid of the may so I need to know if I can wait until you merge the PR or I have to fork and reimplement it myself.

jstedfast:
@Pzixel Your best bet is to fork and implement it yourself, I suspect. May is coming up pretty quick and I don't see 1.9.x coming out before then.

onovotny:
If there's code in master, I'm always happy to publish a 1.9.x prerelease from my fork.

precisemath:
Any ETA as to when the 1.9x release is available?

crypto/test/src/security/test/TestParameterUtil.cs
```diff
@@ -67,7 +67,7 @@ public void TestCreateKeyParameter()
 			Type			expectedType,
 			byte[]			expectedBytes)
 		{
-			Assert.IsTrue(expectedType.IsInstanceOfType(key));
+			Assert.IsTrue(expectedType.GetTypeInfo().IsInstanceOfType(key));
```
nmbro: Can't this just be expectedType.IsIstanceOfType(key)

jstedfast: Depending on the version of NETStandard, some of these methods are on a `TypeInfo` rather than the `Type`. It's something I had to deal with when porting my own libraries to .NETStandard.

crypto/test/src/asn1/test/X509ExtensionsTest.cs
```diff
@@ -100,7 +100,7 @@ public override void PerformTest()
 			}
 		}
 
-		public static void Main(
+		public static void MainOld(
```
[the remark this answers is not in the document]
onovotny: I'm not sure how else to compile this. I get errors of duplicate/ambiguous entry points as the VSTest framework supplies its own Main method in a library. If there's a way to explicitly specify that, I can undo these.

crypto/test/src/security/test/TestParameterUtil.cs
onovotny: This could probably be simplified now that I've removed < NET Standard 2.0. NS 2 should have the full Type surface area.

jstedfast: I was hoping you'd keep .NETStandard < 2.0 because MimeKit targets 1.3, 1.6, and 2.0

onovotny: I can't since the original codebase used Array.Copy that takes a long and that overload doesn't exist before netstandard 2.

I'd strongly suggest nuking 1.3 and 1.6 as literally nothing Microsoft supports targets below 2.0 anymore. 

The lowest supported UWP is 16299 and that's netstandard 2.0. .NET Core 1.0, 1.1, and 2.0 are all out of support. There's .NET 4 multi-target, so all current .net framework's are covered.

jstedfast: Okay, thanks for the info.

crypto/test/src/asn1/test/ASN1IntegerTest.cs
```diff
@@ -60,7 +60,7 @@ public override void PerformTest()
             }
             catch (ArgumentException e)
             {
-                IsEquals("malformed integer", e.Message);
```
nmbro: wouldn't the right way to do this be: `StringAssert.Contains("malformed integer", e.Message);`

version.json
```diff
@@ -0,0 +1,11 @@
+{
+  "version": "1.8.5.2",
+  "publicReleaseRefSpec": [
+    "^refs/heads/master$", // we release out of master
```
nmbro: afaik json doesn't support comments at all

scripts/Sign-Package.ps1
```diff
@@ -0,0 +1,24 @@
...
+
+# Setup Variables we need to pass into the sign client tool
+$appSettings = "$currentDirectory\SignClient.json"
+
+$nupgks = ls $Env:ArtifactDirectory\*.nupkg | Select -ExpandProperty FullName
+
+foreach ($nupkg in $nupgks){
+	Write-Host "Submitting $nupkg for signing"
+
+	.\SignClient 'sign' -c $appSettings -i $nupkg -r $env:SignClientUser -s $env:SignClientSecret -n 'Portable.BouncyCastle' -d 'Portable.BouncyCastle' -u 'https://github.com/onovotny/bc-sharp'
```
nmbro: I think you should prepend this line with a `&`

crypto/test/src/asn1/test/ASN1IntegerTest.cs
onovotny: I agree there's probably a better way to do this, but I was just following what this commit did:

https://github.com/bcgit/bc-csharp/commit/194d05d2d8f4ad5eb2d1d940cb692096f1210485

version.json
onovotny: It does with .NET's parsers.

scripts/Sign-Package.ps1
onovotny: Why?

nmbro: https://ss64.com/ps/call.html
It's the call operator for powershell - sure most of the time it will work without it - but it's just a good habit.

crypto/test/src/asn1/test/ASN1IntegerTest.cs
nmbro: Yeah the entire IsTrue static method also baffles me - I think Bouncycastle is showing some age in it's unit tests at least XD...

johnou:
@rastrup is this on your bat radar?

rastrup:
> @rastrup is this on your bat radar?

Yes it is but why do you ask?

johnou:
@rastrup my team is eager to see .NET Standard 2.0 support in the official distribution.

rastrup:
> @rastrup my team is eager to see .NET Standard 2.0 support in the official distribution.

So am I but I have nothing to do with it and I have not yet seen an official ETA

johnou:
@rastrup wow my bad, serves me right for navigating the web with my phone, I was under the impression that you were one of the project maintainers.

yaakov-h:
Why has this spent 3 years in limbo? Is there a plan to actually incorporate this PR?

johnou:
@peterdettman ptal

jjmanton:
@peterdettman @onovotny Will this PR be merged at some point? As we migrate towards package reference and .net core, I'd love to pull from the official NuGet Package.

ebicoglu:
We are also waiting this PR...

sherland:
Any ETA on this?

jstedfast:
@sherland no ETA - just use the [Portable.BouncyCastle](https://www.nuget.org/packages/Portable.BouncyCastle/) nuget packages.

Kungitoo:
Why no ETA? It would be way better if we could have as much confidence as possible in a cryptography library and that would mean using the portable version under the Bouncy Castle name

John0King:
@ project owner ,  it't been 4 years , and why this pr still not been merge ?  is there anything that stop you to update to Sdk style project ?   if the new visual studio is the blocker , then try visual studio code  !!

darakeon:
@dghgit @peterdettman

flipm0de:
Why is this branch never merged?

John0King:
Same Question :  Why this PR never meged ?    Does the repo owner still active on github ?  Please reply

dghgit:
Just to offer an update, we are now planning a 2.0.0 release which will just provide the platform shift a few weeks from now. We were originally planning to do it earlier, but dealing with all the updates required for TLS took longer than expected and we wanted to get out a final release with those in place as well. Watch this space.

mdsitton:
Just as a follow on here this should be the #1 priority of the library IMO if you check nuget a vast majoirty of users of the library are using other non-offical distributions of the library mainly targeted towards newer version of .net such as .net core or anything that supports .net standard:
https://www.nuget.org/packages?q=BouncyCastle

image

I've left a comment on #50 about further effort that should go into future work where i go over the current minimum supported .net framework versions which shows that .net standard 2.0 should be the minimum development target for this library. I would argue that supporting older .net framework versions and continuing to maintain support for these versions is itself promoting potentially outdated and vulnerable OS's and .net framework versions as these older framework version updates are tied with end of life os's

So because of that i think this PR should be redone with this in mind. and then in the future a BC 2.0 release should be done which drops support for older .net releases. Would be a great time to also do a lot of modernization work to move to generic collections, or Span/Memory based apis potentially. Since it would be a good opportunity for doing an API compatibility break. (though i don't know what BC's policy is on such api deprecation and whatnot.) Older versions can still be maintained in branches for critical bugfixes and security fixes.

clairernovotny:
I'll add that AFAIK, the BouncyCastle nuget package isn't owned/maintained by anyone at BC itself and should not be considered trusted.

My fork is the apparent de-facto package and the CI pipeline is setup and public. The code is signed and published via pipelines, with all reproducible build checks in place, so it's also the most traceable/auditable version with a clear provenance. 

I'm more than happy to continue maintaining/publishing that package as needed, and if the BC team makes a formal request to NuGet, we can get ownership of the main BouncyCastle package transferred over as well.

mdsitton:
Yeah that's fair one thing that's important to me about this issue regardless of who's publishing it, is that the longer this isn't merged the longer it will be before library modernization can be done due to the library targeting .net 1.1 still (as far as i can tell)

Xyncgas:
When will we get .NET 6 support utilizing Span<T> and ref struct

mdsitton:
> When will we get .NET 6 support utilizing Span and ref struct

Looks like now judging by the commits happening in the repo currently!

dghgit:
Yes, we're just waiting on a signing certificate so we can do a proper job on the packaging.

peterdettman:
Here it is: https://www.nuget.org/packages/BouncyCastle.Cryptography.

This release is dedicated to @clairernovotny for keeping the project alive these past few years with `Portable.BouncyCastle`. Although we didn't merge this PR, it was used as the point of departure and to help us figure out the new project formats and tooling. Thanks for all your work, Claire.

PR #71, opened 2017-02-17
Implement HKDF

mtausig:
I have implemented the *HKDFBytesGenerator* class from the java API.
There is a unit test available, but it is a standard nUnit test and does not use BouncyCastle's ITest interface. I hope that is OK.


mtausig:
OK, I obviously missed, that someone has already done that and the code is just missing in the release. D'oh.

peterdettman:
@mtausig Oh, that's unfortunate! We appreciate the intention though, thank you. I assume you needed HKDF for some reason, so if you have any feedback/improvements from using the existing one, they are welcome. Note that until we drop a 2.0 release, we're still at .NET 1.1, and new unit tests should typically be runnable via both nUnit and ITest.

PR #86, opened 2017-08-13
PrivateKeyFactory

doomkin:
Problem solved: cast type error "Org.BouncyCastle.Asn1.DerInteger" to "Org.BouncyCastle.Asn1.DerOctetString".


peterdettman:
I've updated PrivateKeyFactory to also accept INTEGER encodings. This also came up in bc-java: https://github.com/bcgit/bc-java/pull/216 .

PR #92, opened 2017-08-25
Implement Raw Public Keys (Work In Progress)

jimsch:
I really needed to get this running for testing purposes, so I have done a first cut of an implementation.  At this point I know there are server things that still needed to get fixed, but I would like to get a review of the overall direction and perhaps suggestions on ways that would make it better.

I have created an AbstractCertificate class and changed all of the notify functions to use this.  I am not sure this is the best answer, but trying to decide how to deal with some things go here and some things go there did not make a lot of sense.

I have tested the DTLS versions at this time, because that is what I wanted, and will be checking the TLS versions next week.  In addition to this I know that I need to change the framework back.  

One of the changes that still needs to be done is to not send back the client certificate type item if there is not going to be a certificate request of the client.  This means changing the location of the function call to get that information and I am not sure if there are going to be any problems with information not getting populated before that is done as this needs to be done during the ServerHello message creation.


peterdettman:
Apologies for ignoring this so long. I ended up porting raw public key support from bc-java after a PR there: https://github.com/bcgit/bc-csharp/commit/ed081e3fe9634391ac496bd79193a7d00dfa6f07 .

PR #100, opened 2017-11-22
BouncyCastle.csproj: Added trailing slash to OutputPath for compatibility with MSBuild v2

ROM-Knowledgeware:
Without this fix, MSBuild will name the output file 'net20BouncyCastle.dll'.
This change is 100% compatible with newer versions of MSBuild.
p.s. When setting the output path with the Visual Studio GUI (any version I tested) a trailing slash is being used.


peterdettman:
Merged, thanks.

PR #101, opened 2017-11-26
CMSSignedDataGenerator: Added useDerEncodingForCerts parameter to Generate method

ROM-Knowledgeware:
This pull requests add an optional parameter to the Generate() methods that allows switching the default BER format to DER.

Backstory: I needed to have the certificates in the CMS in DER format instead of BER for compatibility reasons, and found out it couldn't be done without copying hundreds of lines of code from BC, this option simplifies the process significantly. I think it should be part of the API and applied to the JAVA version as well.

### DIFF
--- a/crypto/src/cms/CMSSignedDataGenerator.cs
+++ b/crypto/src/cms/CMSSignedDataGenerator.cs
@@ -447,6 +447,17 @@ namespace Org.BouncyCastle.Cms
             return Generate(content, false);
         }
 
+		/**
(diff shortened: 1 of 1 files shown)


peterdettman:
I've added separate `UseDerForCerts` and `UseDerForCrls` properties to `CmsSignedGenerator`, of which `CmsSignedDataGenerator` is a subclass and will respect the setting during `Generate`.

Issue #103, opened 2017-12-11
Error reading the private key of GOST 2012


doomkin:
I generated a private key
```bash
openssl genpkey -algorithm gost2012_256 -pkeyopt paramset:A -out key.pem
```
I tried to read private key from **key.pem** as string
```C#
            using (var keyPemTextReader = new StringReader(pemContent))
            {
                var keyPemReader = new PemReader(keyPemTextReader);
                var pemObject = keyPemReader.ReadObject();
                return pemObject as AsymmetricKeyParameter;
            }

```
As a result, I found out that **bc-csharp** does not support GOST 2012 private keys.
```
Org.BouncyCastle.OpenSsl.PemException : problem creating  private key: Org.BouncyCastle.Security.SecurityUtilityException: algorithm identifier in key not recognised
```
Does anyone plan to port it from **bc-java**?

bcgit:
support for this is now added together with the signature algorithm,

PR #104, opened 2017-12-17
Added implementation of private and public keys GOST R 34.10-2012.

doomkin:
It solves Issue #103


pmpon:
Hi, any chance the request will be accepted?
As you may know GOST R 3410-2001 is outdated and should be replaced with 2012 version till 2019-01-01.

bcgit:
Hi, we ended up taking a slightly different take on this by adding a parameters object. The work for this is now committed together with a signature implementation. Let us know how you go with it.

PR #109, opened 2018-02-06
Adding compatibility with Pkcs5S2 certificates with HMAC256

jperlope:
Do not use the default HMAC with default SHA1 digest anymore. We use digest indicated in pbeParameters.Prf.Algorithm. Tested with HMAC/SHA1 and HMAC/SHA256 on Pkcs5S2 certificates.


jperlope:
I want to write unit tests to verify this fix but I'm only have certificates Pkcs5S2 with HMAC256 from my organization. And I can't publish them. Could anybody tell me how to create a certificate Pkcs5S2 with HMAC256 using BC libraries, please ?

PR #114, opened 2018-03-24
if the parameter "inOff" in Decrypt() is not zero, it will be wrong

ZZMarquis:
if the parameter "inOff" in Decrypt() is not zero, it can not be decrypted as excepted


peterdettman:
Patch applied, thanks.

PR #117, opened 2018-04-09
Supply OrderN and CofactorH to constructor of FpCurve if provided

Devvox93:
FpCurve has a constructor where the order N and cofactor H can be supplied. This pull request uses this constructor when the parameters are supplied in the Asn1Sequence (count == 4).
For the F2mCurve it is still a TODO (that was there already), but my knowledge of crypto (and my time) is currently not sufficient to implement that as well.

I made this because I needed it (redefined the class in my own project) and I thought to share it and improve BC. I hope this is added in a next release.


peterdettman:
We updated this along the way somewhere already; processing would usually begin from the Org.BouncyCastle.Asn1.X9.X9ECParameters class, which now extracts the order/cofactor if available.

PR #118, opened 2018-04-10
ECDSA Plain support

Devvox93:
I've added support for the Plain ECDSA signature algorithms. The main difference for this is that the signatures for this are not DER encoded.
It's not a big difference, but wasn't supported yet and therefore I lost a lot of time to research for this.
Therefore, I've decided to share my work to improve the library :)

I tested it with verifying signatures. Also, because of the default value in the constructor parameters, existing code using the DsaDigestSigner will not break.

If there are any improvements I can make (or maybe did something wrong), I'm open to all feedback :)

### DIFF
--- a/crypto/BouncyCastle.csproj
+++ b/crypto/BouncyCastle.csproj
@@ -51,6 +51,7 @@
     <Compile Include="bzip2\src\CBZip2OutputStream.cs" />
     <Compile Include="bzip2\src\CRC.cs" />
     <Compile Include="src\asn1\BERBitString.cs" />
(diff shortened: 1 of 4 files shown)


peterdettman:
We've finally added plain ECDSA support, along with the BSI OIDs. Code was based on porting from our Java build rather than merge this PR. Actually there are some issues with the PR's PlainEncode and PlainDecode methods, so I would advise switching to the updated BC once 1.8.4 is out with our changes.

Thanks for the suggestion and the sample code. Please let us know if you find any problems.

PR #125, opened 2018-05-21
transplant SM4 algorithm from [https://github.com/bcgit/bc-java]

ZZMarquis:
migrate SM4 algorithm from [https://github.com/bcgit/bc-java]


peterdettman:
Apologies, SM4 was ported recently without reference to this PR.

Issue #140, opened 2018-08-29
What is the official NuGet package?


setycz:
Hello,

if I understand it correctly, this is the official repository for http://www.bouncycastle.org/csharp/ project, right? I would like you to state what is the official NuGet package for this project. Right now, you can find:

- https://www.nuget.org/packages/BouncyCastle/
- https://www.nuget.org/packages/BouncyCastle.Crypto.dll/
- https://www.nuget.org/packages/BouncyCastle.Crypto

each one with different authors. 

I mean that especially for the library with all sorts of security algorithms the distribution should be also secure. Now, it looks like that someone is trying to abuse this library to distribute his (potentially dangerous) code.

setycz:
Thank you @jstedfast , I have opened a PR in which I added this information into README.md so it's available for everyone else.

toasteur:
is there any rule on how any new release translate to an update of the package on nuget.org since the nuget version is still on 1.8.2?

onovotny:
I'm maintaining what can be called the "official" unofficial .NET Standard fork:
https://www.nuget.org/packages/Portable.BouncyCastle/

PR has been sitting here for nearly two years: https://github.com/bcgit/bc-csharp/pull/68

jstedfast:
@onovotny Just want to say thanks for your efforts, they are much appreciated by me and I'm sure many others. Probably not news to you, but thought I'd say it anyway :)

johnhargrove:
Agreed, thank you very much for supporting .NET Standard.

setycz:
@jstedfast , should I add @onovotny 's link to https://www.nuget.org/packages/Portable.BouncyCastle/ to the readme in the PR as well and an official nuget package for .NET Standard version of the library? 

How can I help to get that PR merged?

jstedfast:
I'm not a maintainer, you'd have to ask @peterdettman

bjguillot:
Is there any update on this?  Our project uses BouncyCastle and BouncyCastle.NetCore NuGet packages, but https://www.nuget.org/packages/BouncyCastle/ hasn't been updated since 1.8.6.1, and https://www.nuget.org/packages/BouncyCastle.NetCore/ is more recent, but still is back at 1.8.8

clairernovotny:
Use Portable.BouncyCastle. I'm keeping the fork alive.

peterdettman:
As of just now, the official package is https://www.nuget.org/packages/BouncyCastle.Cryptography .

PR #141, opened 2018-08-30
Adding the list of official NuGet packages into readme

setycz:
Especially for the library with all sorts of security algorithms the distribution should be also secure. Now, there are many unofficial NuGet packages and it can look like that someone is trying to abuse this library to distribute his (potentially dangerous) code. For more info see #140 .


johnhargrove:
I would recommend adding the portable version of the library to the official package as well. Per https://github.com/bcgit/bc-csharp/issues/140#issuecomment-430445962

peterdettman:
As of just now there is just one official package: https://www.nuget.org/packages/BouncyCastle.Cryptography .

PR #151, opened 2018-10-05
Add a new constructor which accepts RsaPrivateKeyStructure

fedelippo:
This commit adds a new constructor which allows to create an instance of
RsaPrivateCrtKeyParameters directly from a RsaPrivateKeyStructure,
rather then building it manually.


peterdettman:
Merged, thanks for the PR. (Takes a few days to appear here though).

PR #152, opened 2018-10-05
Update a documentation comment in Build method

fedelippo:
CertHolder param is present in the Java class method but no longer in
the Csharp one where the X.509 certificate is passed instead.


peterdettman:
Merged, thanks.

PR #178, opened 2019-03-14
Changed util classes from abstract to static

bb-froggy:
A couple of classes in the Utilities namespace are abstract. They have no abstract methods, though, and only static methods. In the internal code, they are never instantiated or inherited.

Thus, the static qualifier is much more appropriate than the abstract qualifier. I changed this in this pull request.

I guess this would be a breaking change if someone used the code strangely, i.e. instantiated one of those classes. I see no reason why someone would do that, though.


peterdettman:
Static classes are a .NET 2.0 language feature, while the code is still compiled for .NET 1.1. That will change soon hopefully, so I'll revisit this at that time.

peterdettman:
These have all been changed now.

PR #185, opened 2019-04-09
Remove the digest from OID for the digestEncryptionAlgorithm

bb-froggy:
The CMSSignedDataGenerator allows adding signers in two ways. One are the methods ```AddSigner```, which work for me, while the other ```AddSignerInfoGenerator``` adds the wrong OID for the signature algorithm to the PKCS#7. In my case, it adds sha256WithRSAEncryption (1 2 840 113549 1 1 11), where other common CMS signed data files I have looked at contain only rsaEncryption (1 2 840 113549 1 1 1).

As per [RFC 2315, Section 9.2](https://tools.ietf.org/html/rfc2315#section-9.2), the "digestEncryptionAlgorithm identifies the digest-encryption algorithm (and any associated parameters) under which the message digest and associated information are encrypted with the signer's private key." I understand this also to say that the digest should not be part of the OID.

With this fix, the correct OID appears in the CMS, and I can use the CMS with other crypto libraries, too.

It would be very kind if a maintainer could review this change.

### DIFF
--- a/crypto/src/cms/CMSSignedDataGenerator.cs
+++ b/crypto/src/cms/CMSSignedDataGenerator.cs
@@ -88,7 +88,8 @@ namespace Org.BouncyCastle.Cms
                 this.sigCalc = sigCalc;
                 this.signerIdentifier = signerIdentifier;
                 this.digestOID = new DefaultDigestAlgorithmIdentifierFinder().find((AlgorithmIdentifier)sigCalc.AlgorithmDetails).Algorithm.Id;
-                this.encOID = ((AlgorithmIdentifier)sigCalc.AlgorithmDetails).Algorithm.Id;
+                this.encOID = Helper.FindEncryptionAlgOID((AlgorithmIdentifier)sigCalc.AlgorithmDetails);
+                
                 this.sAttr = sAttr;
                 this.unsAttr = unsAttr;
                 this.baseSignedTable = baseSignedTable;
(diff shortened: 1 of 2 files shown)


jimsch:
Using rsaEncryption is no longer considered to be the correct answer and is only for PKCS #7 and not for CMS.   Please do not make this change.

bb-froggy:
Ah, thank you for pointing out that I referred to PKCS#7 and that this class is for CMS. I just falsely assumed that this would make no difference.

Anyway, ```AddSigner``` adds the rsaEncryption OID if you use an RSA key. ```AddSigner``` and ```AddSignerInfoGenerator``` should behave consistently, so maybe change ```AddSigner``` instead?

The [CMS RFC describes in Section 5.3](https://tools.ietf.org/html/rfc5652#section-5.3) the signatureAlgorithm, but is not really clear on whether the digest algorithm should be part of it. [Section 10.1.2](https://tools.ietf.org/html/rfc5652#section-10.1.2) describes it more explicit, though, and says

"The SignatureAlgorithmIdentifier type identifies a signature algorithm, and it can also identify a message digest algorithm. Examples include RSA, DSA, DSA with SHA-1, ECDSA, and ECDSA with SHA-256."

This makes it clear in my opinion that rsaEncryption is possible, but it indicates that sha256WithRSAEncryption is also valid.

It would still be good to use rsaEncryption for backwards compatibility with PKCS #7. At least add some way to control which OID is added. I also know at least one CMS library that cannot handle sha256WithRSAEncryption: the class SignedCms in the System.Security.Cryptography.Pkcs package. This library is in preview and has a couple more bugs, so I guess it's not a good reference.

PR #190, opened 2019-05-03
Update Ed448phSigner.cs

southernsun:
typo


peterdettman:
Thanks, but the existing spelling is just the British spelling. The codebase isn't standardi(s|z)ed on a particular English variant.

PR #194, opened 2019-05-09
Update TlsProtocol.cs

deniszykov:
added `TlsProtocol.OfferInput()` overload which specify _offset_ and _length_ of input buffer like in java source https://www.bouncycastle.org/docs/tlsdocs1.5on/org/bouncycastle/tls/TlsProtocol.html


peterdettman:
Merged with minor changes, thanks.

PR #195, opened 2019-05-13
Update GeneralName.cs

rastrup:
Passing through parameter in GetInstance of GeneralName


peterdettman:
I don't think that's an error, since GeneralName is an ASN.1 CHOICE type, and there's some rule that those must always be explicitly tagged. If you are encountering a specific error case please give details on that.

rastrup:
No there is no specific code issue, the correction was found by code inspection of unused parameters and it seemed it needed to be passed through (or removed)

bcgit:
Yes, the value has to be true. The only change that would make sense here is to check the parameter and if it's not true throw an exception but I think it might be a bit late to do that - the reason for the "unused" parameter is so the getInstance() tagged looks like all the others, not something a code analyser would appreciate.

PR #197, opened 2019-05-13
Update DistributionPointName.cs

rastrup:
Passing through parameter in DistributionPointName


peterdettman:
Also a CHOICE type.

peterdettman:
Not a bug (see also https://github.com/bcgit/bc-csharp/pull/195).

PR #201, opened 2019-05-23
Update Arrays.cs

rastrup:
Changed to string since char in StringBuilder constructor gets converted into int for capacity


peterdettman:
Merged, nice catch.

PR #207, opened 2019-06-26
add "cancel" function in DTLS process

fm-DmitryShyryayev:
Added "cancel" feature for DtlsClientProtocol and DtlsServerProtocol in tls layer


avenema:
Would love to see feedback on this. There doesn't seem to be another way that I can see to abort an active DTLS handshake.

peterdettman:
Changes from https://github.com/bcgit/bc-java/pull/549 have now been ported to bc-csharp.

PR #209, opened 2019-07-02
First pass at ChaCha20Poly1305 as an IAeadStreamCipher

LiamPowerMatrikon:
Here's a first pass at adding ChaChaPoly as an IAeadStreamCipher.


peterdettman:
We did end up implementing this, though I'm a bit foggy on the history (Org.BouncyCastle.Crypto.Modes.ChaCha20Poly1305).

PR #217, opened 2019-11-14
configure Signer initialization for Gost2012 algorithms

Skamaev:
Hi, faced some issue with Gost-2012 signature verification.
For example for such signature: [signature_with_file.zip](https://github.com/bcgit/bc-csharp/files/3845325/signature_with_file.zip)

If I call `SignerInformation.Verify(signerCert)` it throws [SecurityUtilityException ](https://github.com/bcgit/bc-csharp/blob/f18a2dbbc2c1b4277e24a2e51f09cac02eedf1f5/crypto/src/security/SignerUtilities.cs#L563)
"1.2.643.7.1.1.2.2WITH1.2.643.7.1.1.1.1 not recognised"

Bouncy castle doesn't know anything about Gost2012 signature algorithm widespread oids (1.2.643.7.1.1.1.1, 1.2.643.7.1.1.1.2), so it can't create DigestSigner object with correct signer and digest at `SignerUtilities.GetSigner(...)`

In PR I add this correct mappings.

By the way, if such configuration can be done by library user, It would be perfect:)

### DIFF
--- a/crypto/src/cms/CMSSignedHelper.cs
+++ b/crypto/src/cms/CMSSignedHelper.cs
@@ -4,17 +4,15 @@ using System.Collections;
 using Org.BouncyCastle.Asn1;
 using Org.BouncyCastle.Asn1.CryptoPro;
 using Org.BouncyCastle.Asn1.Eac;
-using Org.BouncyCastle.Asn1.Iana;
-using Org.BouncyCastle.Asn1.Misc;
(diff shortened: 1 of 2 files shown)


ibetin:
This is a very useful one, please merge asap

scryptan:
bump

peterdettman:
These entries have finally been added (with improvements to the SignerUtilities changes).

PR #235, opened 2020-04-08
Add ability to use Little Endian mode.

buddybu:
So, I added an additional constructor to the BlowfishEngine to allow it to run in a "compatibility" mode, it might be better as a parameter to init, and you can make that suggestion if needed.


buddybu:
I didn't like overriding the constructor, so I removed it.  I renamed the flag to "useLittleEndian" and added a property to allow it to be set.

PR #257, opened 2020-07-05
Fixes case in DTLS handshake where throwing a TlsFatalAlert exception gets silently swallowed

sipsorcery:
The loop running the DTLS handshake logic swallows `IOExceptions` on the premise that they are network timeouts. A problem occurs if a handshake fails (in my case due to `ssl_error_no_cypher_overlap`) and a `TlsFatalAlert` exception is thrown. Because it derives from `TlsException : IOException` it also gets silently swallowed and the DTLS handshake logic attempts to send the outstanding record forever.

This PR adds an exception for `TlsFatalAlerts` to inform the application the DTLS handshake has failed.

### DIFF
--- a/crypto/src/crypto/tls/DtlsReliableHandshake.cs
+++ b/crypto/src/crypto/tls/DtlsReliableHandshake.cs
@@ -117,7 +117,7 @@ namespace Org.BouncyCastle.Crypto.Tls
                         }
                     }
                 }
-                catch (IOException)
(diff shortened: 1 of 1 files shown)


peterdettman:
We've released 1.8.7 which includes a fix for the infinite loop (ported from bc-java version), plus support for an overall handshake timeout.

sipsorcery:
Closed as no longer required.

PR #274, opened 2020-10-17
Test Vectors for EC Point Multiply

BlackthornYugen:
Test multiply with the test vectors downloaded from:
http://point-at-infinity.org/ecc/nisttv


BlackthornYugen:
Oh, I should probably extend SimpleTest like the other classes in that folder shouldn't it?

peterdettman:
The current codebase needs to compile for .NET 1.1. Please avoid named parameters, var keyword, yield keyword, generics. Otherwise looks good.

peterdettman:
Thanks for the PR and making those changes. Merged.

PR #279, opened 2020-11-29
DTLS fixes

niuk:
* Change the size of the `record` buffer in `DtlsRecordLayer.Receive` to be able to hold the actual record. Before, it was sized to fit the decrypted plaintext plus the header, whereas the ciphertext is what's actually sent over the underlying transport.
* `Reset` the `GCMBlockCipher` even when the MAC is invalid, because DTLS needs to be able to decrypt each record independently. Without the reset, corrupt packets can cause the `GCMBlockCipher` object to enter a state in which it will fail to decrypt all subsequent records.
* Ignore `bad_record_mac` "fatal" errors in `DtlsTransport.Receive`, because DTLS is meant to be able to ignore corrupt records.

### DIFF
--- a/crypto/src/crypto/modes/GCMBlockCipher.cs
+++ b/crypto/src/crypto/modes/GCMBlockCipher.cs
@@ -473,15 +473,15 @@ namespace Org.BouncyCastle.Crypto.Modes
             this.macBlock = new byte[macSize];
             Array.Copy(tag, 0, macBlock, 0, macSize);
 
-            if (forEncryption)
+            try
             {
-                // Append T to the message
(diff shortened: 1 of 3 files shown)


peterdettman:
I've implemented the ignoring of corrupt records a little bit differently, and there is now a property `TlsPeer.IgnoreCorruptDtlsRecords` to control whether it is enabled (default false to preserve existing behaviour).

I don't really understand the part about the buffer size, so I didn't make that change. It may just be that the code has evolved since you first proposed it, but feel free to open a new issue if you still see a problem in the latest code.

PR #296, opened 2021-03-04
Fix issue where not providing a TSA policy OID throws a null reference exception

### LINKED ISSUE #295: Not providing a TSA policy ID when generating a timestamp token throws a NullReferenceException
The TSA policy ID is optional per RFC 3161. If you try to generate a timestamp token with a null TSA policy ID, the TimeStampTokenGenerator throws a NullReferenceException on line 56 below.

https://github.com/bcgit/bc-csharp/blob/93b32a75656955faf7996d0f3e0ed391968d2ac6/crypto/src/tsp/TimeStampTokenGenerator.cs#L47-L56

If a DerObjectIdentifier object is not created and instead null is passed in (like in the code snippet below on line 126), trying to access tsaPolicy.Id will throw a NullReferenceException.

https://github.com/bcgit/bc-csharp/blob/93b32a75656955faf7996d0f3e0ed391968d2ac6/crypto/src/tsp/TimeStampTokenGenerator.cs#L117-L132

ndouthit:
The TSA policy ID is optional per RFC 3161. If you try to generate a timestamp token with a null TSA policy ID, the TimeStampTokenGenerator throws a NullReferenceException.

Fixes #295

### DIFF
--- a/crypto/src/tsp/TimeStampTokenGenerator.cs
+++ b/crypto/src/tsp/TimeStampTokenGenerator.cs
@@ -68,7 +68,7 @@ namespace Org.BouncyCastle.Tsp
 
             this.signerInfoGenerator = signerInfoGen;
             this.digestCalculator = digestCalculator;
-            this.tsaPolicyOID = tsaPolicy.Id;
+            this.tsaPolicyOID = tsaPolicy != null ? tsaPolicy.Id : null;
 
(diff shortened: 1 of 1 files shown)


dghgit:
Doh! Thanks for the fix. Merged with some minor changes.

PR #325, opened 2021-10-07
Fix HKDF related files naming

JakubJatczak:
This PR unifies file name casing with classes names, and fixes the `.csproj` includes to work on case sensitive OS'es.


peterdettman:
I've now changed these filenames, though I'm not sure it matters in the modern project setup.

PR #328, opened 2021-10-13
Missing OtherRevocationInfoFormat in CMSSigned classes

iupsilon:
This PR aims to add support to OtherRevocationInfoFormat in CMSSigned* classes.
The implementation is inspired from Java repo.

I recently had to deal with CAdES-LT baseline and I realized that CrlSet can either contain X509Crl and OtherRevocationInfoFormat objects.
Unfortunately the current bc-csharp doesn't manage this aspect and breaks throwing an InvalidCastException in the CreateCrlStore method if crlSet contains OtherRevocationInfoFormat.

I also added the test method TestCrlAndOtherRevocationInfoFormat in order to verify the correctness of the solution.

### DIFF
--- a/crypto/src/cms/CMSSignedData.cs
+++ b/crypto/src/cms/CMSSignedData.cs
@@ -248,7 +248,20 @@ namespace Org.BouncyCastle.Cms
 
 			return crlStore;
 		}
+		
+		/// <summary>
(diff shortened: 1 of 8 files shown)


peterdettman:
Thanks for the PR, this is now fixed. The code has evolved a bit, but I've made equivalent changes.

PR #358, opened 2022-04-17
Add support ArraySegment<byte>.

pengweiqhca:
https://docs.microsoft.com/en-us/dotnet/api/system.arraysegment-1#remarks


fariborz4:
H

PR #360, opened 2022-05-09
csproj: fix file name casing

dlech:
Commit 196bbb0190164b985580d8c4465f6f10f2306bca changed the casing of two file names but did not update the .cspoj file which breaks building on case-senstiive file systems.


dlech:
Hmm... duplicate of  #325.

Issue #374, opened 2022-08-01
Reason in OCSP RevokedStatus is not optional


harrison314:
Instance `RevokedStatus` in `Org.BouncyCastle.Ocsp` it is not possible to create without a reason code.
Well, he is optional.

peterdettman:
PR merged.

PR #375, opened 2022-08-01
Make reason in RevokedStatus (in Org.BouncyCastle.Ocsp) optional.

harrison314:
See https://github.com/bcgit/bc-csharp/issues/374


peterdettman:
Merged, thanks.

PR #378, opened 2022-08-02
Optimize constructors for Time

harrison314:
Optimize constructors for `Org.BouncyCastle.Asn1.X509.Time` and `Org.BouncyCastle.Asn1.Cms.Time`.

In the original code, `DateTime` was format to string, from string was parsed year and ASN1 time structure.
I removed unnecessary steps from this code, and instead of parsing strings, `DateTime` is used directly.


peterdettman:
Merged, thanks.

PR #380, opened 2022-08-26
RFC9146 DTLS Connection ID support

patagonaa:
This PR aims to add [RFC 9146](https://www.rfc-editor.org/rfc/rfc9146.html) DTLS Connection ID support.

Tasks:
- [x] Server support
- [x] Client support
- [x] AEAD Ciphers (MAC calculation)
- [x] Block Ciphers (MAC calculation)
- [ ] Peer Address Update
  - [ ] Way to signal to the caller if a datagram was valid (cryptographically verified and newer than the last)
  - [ ] accept / discard out-of-order packets from the old address without changing addresses?
- [ ] Session handling (either integrated or example implementation)
  - [ ] Mapping of datagrams to their respective sessions
- [ ] Discard invalid datagrams
  - [x] if cryptographically invalid
  - [ ] if other errors (?)
- [ ] Testing with other Crypto Libraries
  - [x] Mbed TLS
  - [ ] californium/scandium
- [ ] Fix / remove TODOs (`// TODO[cid]`)
- [ ] Code review
- [ ] Unit Tests


patagonaa:
I'm a bit stuck on Section 6 of the RFC 9146 (Peer Address Update):
The issue is that I don’t really know how to signal to the application if a record was valid or invalid.

This is also connected with another thing about the DTLS API that has been bothering me for a while now:

Creating a DTLS Server that can handle multiple connections simultaneously is kind of complicated, because you’ll have to
- create and bind a UDP socket
- keep a list of sessions
- for each packet, decide (by IP/port) if it belongs to an existing session or start a new session
- use a queue-based DatagramTransport for each session so we can forward the incoming packets to the respective session
- handle abandoned sessions
- …

Now, with Connection IDs you have to do some additional things:
- Save the connection ID once it has been determined
- Split each packet (or at least, each packet from a new IP/port) into its records and parse them to determine the connection ID (which there doesn’t seem to be a way to easily do using BouncyCastle, as the record parsing is part of the DtlsRecordLayer and thus can’t be used separately)
- Find sessions by connection ID if available
- If a packet for a session has an updated IP/port, determine if it is cryptographically valid and newer than previous packets before updating the address

Especially the address update / packet verification is an issue, because (from the application side) the part where you pass the packets to BouncyCastle (DatagramTransport.Receive()) and the part where the packets are verified (DtlsTransport.Receive()) are separate.
As in, there is no “OfferDatagram”-Method or anything similar that could give immediate feedback if the packet offered was valid and could be processed.

There are also a few “//TODO[cid]” comments with questions/issues that need answering/fixing.

patagonaa:
@peterdettman Have you had a look at this yet? We would love to  have this feature implemented, but we won't be able to do that without some support from someone who knows the code better.

peterdettman:
I've begun work on integrating these changes (not by merge since a lot of changes are needed). The main complications are around API breaks (not your fault of course since this was submitted well before v2 went out).

I am a bit dubious about TlsNullCipher not using the DTLSInnerPlaintext. Other ciphers have a TLS 1.3 path for this that is reusable for connection ID format; I think the only reason that TLS 1.3 path is missing in TlsNullCipher is because it is not permitted in TLS 1.3, whereas it is permitted for DTLS 1.2. Thoughts?

patagonaa:
> I am a bit dubious about TlsNullCipher not using the DTLSInnerPlaintext. [...] Thoughts?

Tbh, I haven't really thought about this, but it's probably correct to use DTLSInnerPlaintext in TlsNullCipher as well.

Also, if you haven't already, make sure to look at my `TODO[cid]` comments in the code, because there were some things that weren't clear to me that will likely have to be solved differently.

peterdettman:
Please see https://github.com/peterdettman/bc-csharp/tree/rfc9146 for progress.

I have further changes in DtlsRecordLayer that can get through a handshake, but I haven't quite worked through all the details yet, and there are some thorny issues around limits particularly in the case of retransmission.

peterdettman:
OK, the branch code should be ready to test now. Use new methods on `DtlsTransport` to `Receive(..., Action recordCallback)` and maybe `ReceivePending(..., Action recordCallback)` if you need it (the latter only processes any records in the queue in case a datagram contains multiple records).

`recordCallback` is invoked for each tls12_cid record that meets the "verified" and "newer" requirements from _6. Peer Address Update_. I'll probably make it a little more general later, but I guess this is all you need to test with. Any peer address update would ideally happen during the callback in case subsequent dispatching of the record causes traffic to be sent to the peer before Receive(Pending) returns.

patagonaa:
Thank you so much, I've just had the first successful connection that was continued using a different UDP endpoint than it has been started with! 🎉 

The API still feels a bit hacky, but with the current architecture I don't really see a better way to do the record verification.

Before, what had to be done already to be able to communicate with multiple clients was:

- create and bind a UDP socket
- keep a list of sessions
- for each packet, decide (by IP/port) if it belongs to an existing session or start a new session
- use a queue-based DatagramTransport for each session so we can forward the incoming packets to the respective session

Now, additionally:
- when establishing a session, save its connection id if one has been negotiated
- when there is a packet from a new endpoint (IP+port) and the packet is of type `tls12_cid`, check if I have a session with this cid and add it to the session's queue (otherwise handle it as a new session)
- once the (DatagramTransport) queue item is passed to BouncyCastle, mark the new endpoint as a possible replacement endpoint
- in the DtlsTransport Receive `recordCallback`, replace the endpoint with the previously marked endpoint

I might write an example server and link it here for reference if I have the time.

patagonaa:
I've now added an example server and client that can connect to each other via UDP (and most importantly, can continue sessions with changing endpoints using connection ids and can handle multiple clients at once):

https://github.com/patagonaa/bc-csharp/tree/dtls-server-example/examples/dtls-connection-id

The interesting parts are the `TestPingServer`, as that is where the session handling is happening and the `PingServerSession` as this is where the endpoint update is proposed and confirmed.

This is in no way a production-ready implementation, but it works for testing.
Limitations:
- no cleanup of stale sessions
- stopping the server doesn't work reliably (`UdpClient.ReceiveAsync(CancellationToken)` seems to hang, even if the token is cancelled or the socket is closed)
- logging and exception handling is missing / incomplete
- sending invalid packets can cause exceptions (type `tls12_cid` with a length shorter than the header+cid)

For testing if it works with changing endpoints, the client changes its endpoint after each packet after sending its payload 10 times.


Also @peterdettman right now it seems like you have to return an empty array in `GetNewConnectionID()` if you want to enable connection id support but not use an connection id yourself. That works, but seems kinda unintuitive imho.

peterdettman:
FYI, I still mean to make minor changes to RecordCallback, which should not fundamentally affect the functionality. Then this branch will be merged and a new release is expected by Monday April 17.

peterdettman:
https://github.com/peterdettman/bc-csharp/tree/rfc9146 now has the final small change if tou want to double check it at your end today.

patagonaa:
Updated my [example server](https://github.com/patagonaa/bc-csharp/tree/dtls-server-example/examples/dtls-connection-id) to use/check the new flag and everything seems to work fine, thank you.

peterdettman:
Merged https://github.com/peterdettman/bc-csharp/tree/rfc9146 . Thanks for the original PR and your feedback.

patagonaa:
Great, thank you for implementing this feature.

peterdettman:
@patagonaa I've just published 2.2.1 to NuGet to fix some issues with 2.2.0. The particularly relevant one here is that (unless using the net60 target) TlsBlockCipher connection ID code was incomplete for the case where encrypt_then_mac was negotiated.

PR #397, opened 2022-11-17
Add NuGet links to README

TimothyMakkison:
Nuget lists several packages when searching `BouncyCastle` each with different versions and compatibilities, it is not clear which is the official or current version. The README currently does not link to NuGet, instead the user has to go to your website, scroll down and find the official version in the releases section.

This PR adds widgets linking to the NuGet package and a section demonstrating installing this package from NuGet. Hopefully it will help people migrate and boost the pacakges visibility.


peterdettman:
Merged, thanks.

Issue #354, opened 2022-04-04
Error getting Content from CmsEnvelopedData with AES128-GCM


Nintai:
I got a problem with decrypting `CmsEnvelopedData` that was encrypted with the bouncy castle library in JAVA.

The error occurs when calling `var decryptedData = recipient.GetContent(privKey);` where `recipient` is of type `KeyTransRecipientInformation`.

From what I could find out, the C# library is getting the IV Length from this line in `ParameterUtilities.cs` in the method `GetCipherParameters`:
`int basicIVKeySize = FindBasicIVSize(canonical);`

In the case of `canonical` being "AES128" it will return 16.

As long as `basicIVKeySize` is not -1 it will try to convert the parameter to an octetstring assuming that you only get the IV and not the IV Size:
```
if (basicIVKeySize != -1 || canonical == "RIJNDAEL" || canonical == "SKIPJACK" || canonical == "TWOFISH")
{
    iv = ((Asn1OctetString) asn1Params).GetOctets();
}
``` 

The Parameter inside of the `CmsEnvelopedData`, which was generated with the java library, seems to be:
`{[#1234567890abcdef12345678, 16]}`
which obviously causes an exception when trying to convert it to `Asn1OctetString`, because its a `DerSequence` instead.

So in my opinion there are 2 things that could be the case:
Either 
the java library shouldnt add the IV length when it can be derived from the algorithm
or
the c# library should check if it gets a `DerSequence` and convert it to IV & IV length if thats the case

Should also note that decrypting in java itself works just fine, but thats not what we want to use for decrypting.

peterdettman:
This should now be fixed and will be in the 2.1.0 release, expected shortly.

jumajeric:
Hello!
After the update it seems that the encryption now throws the same error as did the decryption. Is there a correlation?? #475

PR #402, opened 2022-11-29
Fix for Issue #354. "Error getting Content from CmsEnvelopedData with AES128-GCM"

PPeMg:
This is a fix for issue #354 . As I commented on the issue, I have limited knowledge about this so it would be nice if you can review this PR and ensure that this fix doesn't cause any incompatibility with other parts of the library.

### DIFF
--- a/crypto/src/security/ParameterUtilities.cs
+++ b/crypto/src/security/ParameterUtilities.cs
@@ -18,6 +18,8 @@ namespace Org.BouncyCastle.Security
 {
     public static class ParameterUtilities
     {
+        private static readonly IList<string> IvUnsupportedAlgorithms = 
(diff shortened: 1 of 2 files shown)


peterdettman:
This should now be fixed and will be in the 2.1.0 release, expected shortly.

This PR is in the right ballpark, but there is an actual ASN.1 type (`GcmParameters`) associated with this sequence, and also the (optional) second element of the sequence is actually the tag length (i.e. mac size) to use with GCM, so the parameters have to be of type `AeadParameters` in order to be able to properly capture the full information.

Also, for completeness we support all of Aes128Gcm, Aes192Gcm, Aes256Gcm, and the corresponding Ccm parameters too.

PR #410, opened 2022-12-27
Update Asn1GeneralizedTime.cs with more precision

RobLinux:
added 3 more rounds of precision in FromString


peterdettman:
I think it is not so simple. The current class relies on not losing precision in the conversion to DateTime, where a tick is 100ns. So if the value has higher precision, probably ToDateTime() shouldn't be available (maybe there could be other methods that are clear about giving a "nearest" value of some kind, or methods to compare to a DateTime) , and also ToStringCanonical would have to operate on the high-precision string directly and not rely on conversion from DateTime.

RobLinux:
The thing is the current solution is not working with Apple ASN objects. So indeed some precisions would be lost, but at least the objects would be parsed correctly without exception. 

What would you recommend in this case?

RobLinux:
Any update on this ? what can we do ? it's actually a problem in my organization

RobLinux:
Because the problem is that currently I can't deserialie a datetime with a higher precision, which is a problem.

RobLinux:
https://chat.openai.com/share/6471db26-d1aa-4e76-bd69-73206ba70ad4 what about this ?

PR #416, opened 2023-01-25
Fixed cache in DerObjectIdentifier

kmosolov:
In the process of profiling an application that performs the improvement of a large number of signatures to the CAdES-XL format in parallel in many threads, I found that this lock is a "bottleneck".
Therefore, I tried to replace the explicit lock with the use of ConcurrentDictionary, which significantly accelerated the process.
It also turned out that applying the 1023 bit mask to HashCode leads to a large number of cache misses both in the tests of this project and on my real data, so it was removed, without it cache misses are no longer observed, but I did not delete the code handling this situation.

### DIFF
--- a/crypto/src/asn1/DerObjectIdentifier.cs
+++ b/crypto/src/asn1/DerObjectIdentifier.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Concurrent;
 using System.IO;
 using System.Text;
 
@@ -80,7 +81,7 @@ namespace Org.BouncyCastle.Asn1
 
         private const long LongLimit = (long.MaxValue >> 7) - 0x7F;
 
-        private static readonly DerObjectIdentifier[] cache = new DerObjectIdentifier[1024];
+        private static readonly ConcurrentDictionary<int, DerObjectIdentifier> cache = new ConcurrentDictionary<int, DerObjectIdentifier>();
 
         private readonly string identifier;
         private byte[] contents;
@@ -206,18 +207,16 @@ namespace Org.BouncyCastle.Asn1
         internal static DerObjectIdentifier CreatePrimitive(byte[] contents, bool clone)
         {
             int hashCode = Arrays.GetHashCode(contents);
-            int first = hashCode & 1023;
 
-            lock (cache)
+            DerObjectIdentifier entry = cache.GetOrAdd(hashCode, new DerObjectIdentifier(contents, clone));
+
+            if (!Arrays.AreEqual(contents, entry.GetContents()))
             {
-                DerObjectIdentifier entry = cache[first];
-                if (entry != null && Arrays.AreEqual(contents, entry.GetContents()))
-                {
-                    return entry;
-                }
-
-                return cache[first] = new DerObjectIdentifier(contents, clone);
+                entry = new DerObjectIdentifier(contents, clone);
+                cache[hashCode] = entry;
             }
+
+            return entry;
         }
 
         private static bool IsValidIdentifier(string identifier)
--- a/crypto/src/asn1/DerObjectIdentifier.cs
+++ b/crypto/src/asn1/DerObjectIdentifier.cs
@@ -208,7 +208,7 @@ namespace Org.BouncyCastle.Asn1
         {
             int hashCode = Arrays.GetHashCode(contents);
 
-            DerObjectIdentifier entry = cache.GetOrAdd(hashCode, new DerObjectIdentifier(contents, clone));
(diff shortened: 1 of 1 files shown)


crypto/src/asn1/DerObjectIdentifier.cs
```diff
@@ -206,18 +207,16 @@ private byte[] GetContents()
...
-            int first = hashCode & 1023;
 
-            lock (cache)
-            {
-                DerObjectIdentifier entry = cache[first];
-                if (entry != null && Arrays.AreEqual(contents, entry.GetContents()))
-                {
-                    return entry;
-                }
+            DerObjectIdentifier entry = cache.GetOrAdd(hashCode, new DerObjectIdentifier(contents, clone));
```
petarpetrovt, line 211: No point caching it if you parse it each time.
Add `_ =>` in `GetOrAdd`.

```csharp
DerObjectIdentifier entry = cache.GetOrAdd(hashCode, _ => new DerObjectIdentifier(contents, clone));
```

kmosolov: Got lost when transferring from the test branch, fixed it.

harrison314:
I think it is necessary to limit the size of the cache (_ConcurrentDirectory_), without that this is a solution with  _memory leak_ (oids are never released from memory, which can cause problems with long-running applications).

kmosolov:
> I think it is necessary to limit the size of the cache (_ConcurrentDirectory_), without that this is a solution with _memory leak_ (oids are never released from memory, which can cause problems with long-running applications).
> 
> Another option to get rid of the lock is to use `Interlocked.CompareExchange` and `Interlocked.Exchange` on the `cache` array.

It seems to me that the amount of memory occupied by the cache, even with long-term operation of the application, cannot be large, because the number of OIDs is not so large and the size of the cached data is also small.

harrison314:
>It seems to me that the amount of memory occupied by the cache, even with long-term operation of the application, cannot be large, because the ammount of OIDs is not so large and the size of the cached data is also small.

When the OIDs comes to the system from an external environment (eg. web application for certificate/signature validation), such an unlimited cache can be used for a DOS attack by successively sending all valid OIDs.

When I have time, I will try to implement the cache using `Interlocked`, and you can test the performance impact.

harrison314:
@kmosolov try this alternative:

```cs
        internal static DerObjectIdentifier CreatePrimitive(byte[] contents, bool clone)
        {
            int hashCode = Arrays.GetHashCode(contents);
            int first = hashCode & 1023;

            DerObjectIdentifier entry = Cache[first];
            if (entry != null && Arrays.AreEqual(contents, entry.GetContents()))
            {
                return entry;
            }

            DerObjectIdentifier newEntry = new DerObjectIdentifier(contents, clone);

            Interlocked.CompareExchange(ref Cache[first], newEntry, entry);

            return newEntry;
        }
```

or this alternative:
```cs
        internal static DerObjectIdentifier CreatePrimitive(byte[] contents, bool clone)
        {
            int hashCode = Arrays.GetHashCode(contents);
            int first = hashCode & 1023;

            DerObjectIdentifier entry = Cache[first];
            if (entry != null && Arrays.AreEqual(contents, entry.GetContents()))
            {
                return entry;
            }

            DerObjectIdentifier newEntry = new DerObjectIdentifier(contents, clone);

            DerObjectIdentifier originEntry = Interlocked.CompareExchange(ref Cache[first], newEntry, entry);
            if (object.ReferenceEquals(originEntry, entry))
            {
                return newEntry;
            }

            if (originEntry != null && Arrays.AreEqual(contents, originEntry.GetContents()))
            {
                return originEntry;
            }

            return newEntry;
        }
```

peterdettman:
@kmosolov Could you please test this and see how it compares for your scenario:

```
internal static DerObjectIdentifier CreatePrimitive(byte[] contents, bool clone)
{
    int index = Arrays.GetHashCode(contents);

    index ^= index >> 20;
    index ^= index >> 10;
    index &= 1023;

    return Cache.AddOrUpdate(index,
        _ => new DerObjectIdentifier(contents, clone),
        (k, v) => Arrays.AreEqual(contents, v.GetContents())
            ?   v
            :   new DerObjectIdentifier(contents, clone));
}

```

Also, in the original BC code, do the cache misses improve if you just use the first few lines from above (i.e. xor in all the hashCode bits)?

peterdettman:
I went with essentially the second version of the `Interlocked.CompareExchange` approach suggested by @harrison314, and also with mixing in all the `GetHashCode()` bits.

Feedback would be appreciated once we have the 2.1.0 release out, hopefully this week.

PR #420, opened 2023-02-17
Removing unnecessary locks in X509Certificate and X509Crl

harrison314:
Lock with null check is replaced to assign to local variable (is atomic) and null check.

Lock with condition assign is replaced to `Interlocked.CompareExchange`.

### DIFF
--- a/crypto/src/x509/X509Certificate.cs
+++ b/crypto/src/x509/X509Certificate.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Net;
 using System.Text;
+using System.Threading;
 
 using Org.BouncyCastle.Asn1;
 using Org.BouncyCastle.Asn1.Misc;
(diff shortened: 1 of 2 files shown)


peterdettman:
Should return the CompareExchange return value if it's not null; EC public keys, for example, can cache precomputations, so we want maximal sharing.

harrison314:
I don't think it is necessary to check the output value `CompareExchange` if is null, because `Interlocked.CompareExchange(ref publicKeyValue, temp, null);` set `publicKeyValue` to `temp` if  `publicKeyValue` is `null`. 

So if `publicKeyValue` is not `null`, then the value of `publicKeyValue` will not be overwritten.

peterdettman:
Weird, somehow I was reading `return temp`, forget that part.

I think I would still prefer `return CompareExchange(..., null) ?? temp;` though.

harrison314:
@peterdettman  I added `return CompareExchange(..., null) ?? temp;`

peterdettman:
Merged, thanks.

PR #445, opened 2023-03-30
make initial DTLS handshake resend time configurable

patagonaa:
This PR makes the initial time after which a DTLS handshake packet is resent (after getting no response) configurable.
This can be useful to reduce unneccessary retransmissions in networks that are usually reliable, but have a high latency (some mobile networks do) and/or when working with low-power embedded devices that might take a few seconds to run the key exchange algorithms.


peterdettman:
Merged, thanks.

Note that I had to comment out the TlsPeer method declaration (until a next major version) since it technically breaks backward compatibility.

patagonaa:
Great, thank you.

PR #452, opened 2023-04-27
Fixed a comment to refer to correct PKCS

bb-froggy:
The OID is for PKCS#5, but the comment said it was PKCS#12


peterdettman:
Merged, thanks.

PR #453, opened 2023-04-30
Update SPHINCSPlusParameters.cs

nd1012:
FYI: The `SphincsPlusParameters` doesn't implement the `ICipherParameters`, which breaks the API standard that is implemented by most (or all) of the other algorithms.


peterdettman:
No, the cipher parameters for Sphincs+ are either SphincsPlusPrivateKeyParameters for signing (optionally wrapped with ParametersWithRandom) or SphincsPlusPublicKeyParameters for verification. SphincsPlusParameters is more like ECDomainParameters; they are available on the keys, but not suitable for passing to a signer Init.

PR #456, opened 2023-05-03
Update SignerId.cs

gungora:
When there is reference equality, Equals() should return true. I encounter this issue when calling SignerInformationStore.GetFirstSigner().


peterdettman:
Merged, thanks.

PR #472, opened 2023-06-28
Add support ArraySegment<byte>.

pengweiqhca:
Void memory allocate on some time.

``` C#
// var array = new byte[length];
var array = ArrayPool<byte>.Shared.Rent(length); // array.Length maybe greater than length;

Asn1Sequence.GetInstance(new ArraySegment<byte>(array , 0, length));
```


peterdettman:
Closing because we don't really have an interest in this given the availability of span code for most things. Not opposed to a new PR for some specific case where spans don't/can't cut it (and where it matters).

PR #476, opened 2023-07-21
Fix NullReferenceException in PbeUtilities

EnigmaTriton:
Hello,

I stumble upon a `null` when trying to read a PKCS#12 created with OpenSSL using unusual parameters.
Namely, SHA512 was used for the HMAC function.

Adding SHA512 (and SHA384 for completion), was easy enough.
However, I think we might see other similar cases with other unusual parameters that OpenSSL or another tool might generate (the obvious ones I see being SHA512/xxx or SHA3 family).

We probably still need to raise an error but I guess there would be a cleaner one that would indicate the real issue instead of a side effect that happens later (I add both `NullReferenceException` and `ArgumentNullException` depending of other code changes I tried so I’m not sure which one happened with current code).

Can someone help me with the proper handling of the error?

### DIFF
--- a/crypto/src/security/PbeUtilities.cs
+++ b/crypto/src/security/PbeUtilities.cs
@@ -131,6 +131,12 @@ namespace Org.BouncyCastle.Security
             Algorithms["PBEWITHHMACSHA256"] = "PBEwithHmacSHA-256";
             Algorithms["PBEWITHHMACSHA-256"] = "PBEwithHmacSHA-256";
             Algorithms[NistObjectIdentifiers.IdSha256.Id] = "PBEwithHmacSHA-256";
+            Algorithms["PBEWITHHMACSHA384"] = "PBEwithHmacSHA-384";
+            Algorithms["PBEWITHHMACSHA-384"] = "PBEwithHmacSHA-384";
+            Algorithms[NistObjectIdentifiers.IdSha384.Id] = "PBEwithHmacSHA-384";
+            Algorithms["PBEWITHHMACSHA512"] = "PBEwithHmacSHA-512";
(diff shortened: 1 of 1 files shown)


peterdettman:
I suggest throwing ArgumentNullException if the method argument `algorithm` is null. Then if `mechanism` comes back null:

`throw new SecurityUtilityException("Algorithm " + algorithm + " not recognised.");`

EnigmaTriton:
I didn’t found any absence of algorithm in the PKCS#12 I’ve got.

My guess is that we will at least have the OID (that’s what happened in this case) so I just added the check after for "not recognised".

Do you think the preliminary check is required?

peterdettman:
It's probably best to add the null check on algorithm, though it will be thrown by the dictionary lookup anyway.

Also, you can't use the "is null" form since we also target earlier C# versions.

EnigmaTriton:
I was not sure about which was the minimum version of C# supported and I believe that Visual Studio 2013 is still supported by BC but does not support C# 6 where `nameof()` was introduced so I didn’t use it. Just tell me if I missed something.

EnigmaTriton:
Oops, I can see that it’s already in use. I’ll fix it right away.

EnigmaTriton:
Everything looks fine on my side, I opened a bunch of PKCS#12 I had without any issue.
I also ran most of the BouncyCastle test suite (some tests seem to be stuck but running them in debug mode doesn’t go through the modified paths) as well as my application test suite without any error.

Can I help with anything else?

peterdettman:
Merged for upcoming 2.3.0 release, thanks.

PR #507, opened 2023-12-24
Added SRP message verification using RFC2945

jimm98y:
Added methods for calculating M1 using the formula described in RFC 2945. I tried to avoid any breaking changes, which is why the PR just adds new methods while keeping the original implementation unchanged.

### DIFF
--- a/crypto/test/src/crypto/test/SRP6Test.cs
+++ b/crypto/test/src/crypto/test/SRP6Test.cs
@@ -32,20 +32,20 @@ namespace Org.BouncyCastle.Crypto.Tests
 
 	    public override void PerformTest()
 	    {
-	    	rfc5054AppendixBTestVectors();
-			rfc2945MessageVerify();
+            rfc5054AppendixBTestVectors();
+            rfc2945MessageVerify();
 
             testMutualVerification(Srp6StandardGroups.rfc5054_1024);
             testClientCatchesBadB(Srp6StandardGroups.rfc5054_1024);
             testServerCatchesBadA(Srp6StandardGroups.rfc5054_1024);
 
-			testWithRandomParams(256);
-			testWithRandomParams(384);
-			testWithRandomParams(512);
-	    }
+            testWithRandomParams(256);
+            testWithRandomParams(384);
+            testWithRandomParams(512);
+        }
 
         private void rfc2945MessageVerify()
-		{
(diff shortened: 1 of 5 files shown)


crypto/test/src/crypto/test/SRP6Test.cs
```diff
@@ -32,18 +32,69 @@ public override string Name
...
-			testWithRandomParams(256);
-			testWithRandomParams(384);
-			testWithRandomParams(512);
-	    }
+            testWithRandomParams(256);
+            testWithRandomParams(384);
+            testWithRandomParams(512);
+        }
+
+        private void rfc2945MessageVerify()
```
cipherboy, line 47: Do you happen to know if there's any known answer tests floating around for this version of SRPs? Below we have tests for RFC 5054 which include known-answer tests, so presumably you could tell whether or not your given (third-party) SRP implementation corresponds to RFC 5054 by feeding it these vectors... it'd be good to have that for the RFC 2945 implementation as well so we can ensure interoperability!

jimm98y: Searching for RFC 2945 I found this: https://github.com/secure-remote-password/test-vectors/tree/8381017c812a273f30ba866af673e3726b314ab3 Their library seems to implement both RFC 5054 and RFC 2945 - knowing this before, it might have simplified the task quite a bit :)
I can just confirm that my implementation seems to work fine with Apple's after more than 2 months of testing. I might be able to capture some test vectors as well.

cipherboy: \o hello @jimm98y -- I'm not too particular about the source of the vectors, just that there's something we can potentially point to and say, we expect to be compatible with these values; check and see if your SRP protocol matches and if not, we'll need to potentially implement a third. :-) 

That said, if those vectors pass with this code, happy to take them!

jimm98y: @cipherboy Now they do! It turned out I've made a mistake porting my original code to BC: I did not realize that HashPaddedTriplet does exactly what the name suggests - it pads to the desired length and then hashes. For RFC2945 M2 the padding must not be performed for M1 and K components, so I had to add a new method CalculateM2RFC2945 that implements a customized hash function.

PR #529, opened 2024-03-25
Add SignedPublicKeyAndChallenge into Asn1/Mozilla.

harrison314:
Add SignedPublicKeyAndChallenge ASN1 strusture into Asn1/Mozilla.

Signed SignedPublicKeyAndChallenge  is used for certification requests without a subject name. Although it is no longer supported by Mozilla browsers.


peterdettman:
Merged, with some reworking to keep it more like bc-java.

Note in particular that the Verify method was moved to a class in the Org.BouncyCastle.Mozilla namespace.

Issue #422, opened 2023-03-02
Remove embedded resources when using dotnet trimmer


tranb3r:
I was previously using Portable.BouncyCastle in my app, and I'd like to migrate to BouncyCastle.Cryptography.
However, Portable.BouncyCastle.dll 1.9.0 weights about 3MB, while BouncyCastle.Cryptography.dll 2.1.1 weights more than 6MB, because of a few embedded resources.
In my app I'm using dotnet trimmer to remove useless code, and this works fine with Portable.BouncyCastle.
However, with BouncyCastle.Cryptography, the trimmer does not remove the embedded resources, resulting in the size of my app increasing by more than 3MB.

What are those embedded resources?
Is there any solution to trim them?

Thanks!

peterdettman:
They are the following .bz2 files within the PQC code:

- pqc/crypto/picnic/lowmcL1.bz2 ("Org.BouncyCastle.pqc.crypto.picnic.lowmcL1.bz2")
- pqc/crypto/picnic/lowmcL3.bz2 ("Org.BouncyCastle.pqc.crypto.picnic.lowmcL3.bz2")
- pqc/crypto/picnic/lowmcL5.bz2 ("Org.BouncyCastle.pqc.crypto.picnic.lowmcL5.bz2")
- pqc/crypto/sike/p434.bz2 ("Org.BouncyCastle.pqc.crypto.sike.p434.bz2")
- pqc/crypto/sike/p503.bz2 ("Org.BouncyCastle.pqc.crypto.sike.p503.bz2")
- pqc/crypto/sike/p610.bz2 ("Org.BouncyCastle.pqc.crypto.sike.p610.bz2")
- pqc/crypto/sike/p751.bz2 ("Org.BouncyCastle.pqc.crypto.sike.p751.bz2")

with a total size of more than 3MB. Sike at least will be removed for v3.0 in any case.

If trimming these you probably want to explicitly trim also the entire Org.BouncyCastle.Pqc namespace (there are some factory classes in Org.BouncyCastle.Pqc.Crypto.Utilities that might otherwise cause issues).

tranb3r:
Ok.
However, even the trimmer removes the entire Org.BouncyCastle.Pqc namespace, it does not remove the embbeded resources.
Do you know how to force it?

peterdettman:
Have you tried removing the Org.BouncyCastle.pqc namespace (lowercase `pqc`)?

tranb3r:
[...]

PR #534, opened 2024-04-18
Embed picnic data in source to make it trim-friendly

Rob-Hague:
Move the constant data for picnic from the bz2 files embedded as resources in the assembly to be embedded as static properties in the source. Those properties can then be analysed and trimmed away by the trimmer when unused.

In a test app which generates an ECDH key pair (i.e. does not use picnic), this results in the following changes to the size of BouncyCastle.Cryptography.dll:

|           |  before |   after |
| --------- | ------- | ------- |
| untrimmed | 6989 KB | 6975 KB |
|   trimmed | 3993 KB | 2791 KB |

That is, the untrimmed assembly size is unaffected with this change, but the trimmed assembly is reduced by 1202 KB which is exactly the size of the bz2 files.

Of the remaining 2791 KB, 2229 KB is the sike bz2 data, which is not treated here.

The source code was generated by reflecting over instances of the existing classes:

```c#
using Org.BouncyCastle.Pqc.Crypto.Picnic;
using System;
using System.CodeDom.Compiler;
using System.IO;
using System.Linq;
using System.Reflection;

LowmcConstantsL5 lc = new();

FieldInfo[] fields = lc.GetType().GetFields(BindingFlags.NonPublic | BindingFlags.Instance);

using StreamWriter sw = new(@"C:\tmp\LowmcConstants.txt");
using IndentedTextWriter tw = new(sw);

tw.Indent++; //namespace
tw.Indent++; //class
tw.WriteLine(); // to get tabs on next line

foreach (FieldInfo uintField in fields.Where(f => f.FieldType == typeof(uint[])))
{
    uint[] data = (uint[])uintField.GetValue(lc);

    int nonZeroDataLength = data.AsSpan().LastIndexOfAnyExcept(0u) + 1;

    tw.WriteLine($"/// <summary>Length: {nonZeroDataLength}</summary>");
    tw.WriteLine($"private static readonly uint[] s_{uintField.Name} = new uint[]");
    tw.WriteLine("{");
    tw.Indent++;

    foreach (uint[] chunk in data.Take(nonZeroDataLength).Chunk(8))
    {
        tw.Write("0x");
        tw.Write(string.Join(", 0x", chunk.Select(u => u.ToString("X8"))));
        tw.WriteLine(",");
    }

    tw.Indent--;
    tw.WriteLine("};");
    tw.WriteLine();
}
```

Contributes to #422


peterdettman:
Merged, thanks for the PR. If anyone wants to do something similar for Sike it would be welcome (it will be removed completely in 3.0, but that's probably still a ways off).

Rob-Hague:
Great, thanks. I'll give it a go - it seems slightly more complicated but still viable.

Rob-Hague:
I've got the change at https://github.com/bcgit/bc-csharp/compare/master...Rob-Hague:bc-csharp:sike

It's currently based on GitHub/master so doesn't include this change and fails the picnic tests because it removes `<EmbeddedResource Include="pqc\crypto\**\*.bz2" />` from the .csproj. Should I submit it against GitHub/master, this branch or wait for GitHub/master to update (or you can directly cherry-pick my branch)?

Presumably there are no conflicts whichever way.

Thanks

PR #545, opened 2024-06-21
Fix a wrong delete

tohidemyname:
If the returnKey size is more than one, the first deletion will change the index of returnKey.*. The follow-up deletions will delete the  wrong items.


peterdettman:
I see that the removals will cause the check on the "next" entry to be skipped, but I don't see how it could delete wrong items.

In any case, the fix looks good, thanks for reporting the issue.

peterdettman:
Fixed in 2.5.1.

PR #547, opened 2024-06-21
fixing another wrong delete in PgpPublicKey

tohidemyname:
Just notice that this class has another similar bug


peterdettman:
Maybe not a bad change to make, but I don't see an error here - idSigs list is not being modified, only a list inside idSigs[i].

PR #520, opened 2024-02-10
Initial implementation of Argon2 (ported from bc-java)

roughconsensusandrunningcode:
Port of Argon2BytesGenerator, Argon2Parameters and  ICharToByteConverter/PasswordConverter from bc-java, with unit tests.

Issue #202, opened 2019-05-23
Argon2 Support


sopelt:
Hi, now that the Java Bouncy Castle has added support for Argon2 (which should be the go to password hashing algorithm for new systems according to best practices like https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Password_Storage_Cheat_Sheet.md#leverage-an-adaptive-one-way-function ) are there any plans (or interest) in adding it to bc-csharp as well?

Kind regards

peterdettman:
Merged https://github.com/bcgit/bc-csharp/pull/520 which resolves this.

PR #559, opened 2024-08-22
Argon2

zer0x64:
## Describe your changes

Added support for the Argon2 password-based key derivation function.

## How has this been tested?

Right now, I implemented tests for the test cases in the RFC. I can add more test cases if needed.

- [x] This change requires a documentation update
- [x] Remove unnecessary indirection
- [x] Parameters checking
- [x] Memory zeroing
- [x] Multithreading
- [x] Add more tests
- [x] Optimizations using spans

### DIFF
--- a/crypto/src/crypto/util/OpenSshPrivateKeyUtilities.cs
+++ b/crypto/src/crypto/util/OpenSshPrivateKeyUtilities.cs
@@ -162,13 +162,24 @@ namespace Org.BouncyCastle.Crypto.Utilities
                     if (sequence[3] is Asn1TaggedObject && sequence[2] is Asn1TaggedObject)
                     {
                         ECPrivateKeyStructure ecPrivateKey = ECPrivateKeyStructure.GetInstance(sequence);
-                        DerObjectIdentifier curveOID = DerObjectIdentifier.GetInstance(ecPrivateKey.Parameters);
-                        X9ECParameters x9Params = ECNamedCurveTable.GetByOid(curveOID);
-                        result = new ECPrivateKeyParameters(
-                            ecPrivateKey.GetKey(),
-                            new ECNamedDomainParameters(
-                                curveOID,
-                                x9Params));
+
+                        X962Parameters parameters = X962Parameters.GetInstance(ecPrivateKey.Parameters.ToAsn1Object());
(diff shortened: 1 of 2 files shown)


zer0x64:
Related issue:
https://github.com/bcgit/bc-csharp/issues/202

zer0x64:
Putting this as a draft right now. I got the RFC test cases working, but I still have to cleanup my code, optimize it and add multithreading support. I just wanted to open as a draft ASAP to be able to address requested changes.

One thing that must be brought to attention is multithreading support. I'd assume we want to make multithreading support optionnal, for portability, but I was wondering how you'd like that to be implemented? Via an additionnal parameter on the API, or via a preprocessor directive?

On thing I'd also like to fix before merging is using Span and MemoryMarshal to reference byte[]<->ulong[] arrays directly instead of copying when on a little endian machine. Another thing is that I don't support memory clearing yet, but that should be easy to add.

zer0x64:
Ready for review! A couple of notes:

1. Right now, multithreading is enabled if there is more than one lane. Do we want to put an (opt-out) argument or preprocessor directive to disable multithreading support?
2. Technically, since Blake2b works on QWORDs too, it might be possible to optimize for big endian and framework platforms by implementing and using a QWORDs based api for when Blake2b is used. However, this is not used for the intensive part of the algorithm so I don't think it's too important.
3. Do we want to add a user-facing `Span<byte>` API for netcore3.1+ users? That could be work for a future PR.
4. My `Zeroize<T>` generic function might be a better fit in some utils class, like an Arrays.Clear() overload

peterdettman:
I just merged an earlier Argon2 PR: https://github.com/bcgit/bc-csharp/pull/520 . I see there are a few diffs in your PR (e.g. span-related and tasks) and will probably incorporate those.

zer0x64:
Great news! I don't mind opening up a new PR to add those improvements myself, just tell me if you'd rather have me do it or do it yourself!

peterdettman:
@zer0x64 A PR adding improvements would be welcome.

zer0x64:
This PR got automatically closed because I've put the work on a branch on my fork instead of master, but the code is still over there. I'll open a new PR to add the requested changes

PR #560, opened 2024-08-27
Additional Pkcs10CertificationRequest constructor for interoperability

bencemali:
Creating a `Pkcs10CertificationRequest` instance with key types not handled by `SubjectPublicKeyInfoFactory.CreateSubjectPublicKeyInfo` require heavy workarounds, and adding a constructor taking an already constructed `SubjectPublicKeyInfo` would help facilitate pqc/custom keys in `Pkcs10CertificationRequest` CSRs.

### DIFF
--- a/crypto/src/pkcs/Pkcs10CertificationRequest.cs
+++ b/crypto/src/pkcs/Pkcs10CertificationRequest.cs
@@ -286,12 +286,41 @@ namespace Org.BouncyCastle.Pkcs
             Init(signatureFactory, subject, publicKey, attributes);
         }
 
+        /// <summary>
+        /// Instantiate a Pkcs10CertificationRequest object with the necessary credentials.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


peterdettman:
Merged, thanks for the PR.

Sidenote: the recently standardized PQC algorithms will be getting integrated into the regular factories/registries soon.

bencemali:
Thanks @peterdettman! Are there any efforts to support pqc hybrids with ec/rsa keys also in the regular factories/registries? I am thinking of a general `HybridKeyParameters`, not a separate implementation for all the different combinations. I guess at first only the standardized algorithms would have to be accounted for in these hybrids, but then it could be expanded. OIDs for the hybrids could be imported from [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) for example. If there's nothing yet I could provide a PoC in a few days.

PR #565, opened 2024-09-11
PQC hybrid key support

bencemali:
This PR brings (preliminary) support for hybrid post-quantum + elliptic curve KEMs and signatures. The hybrid serialization formats broadly follow the one used in [oqs-provider](https://github.com/open-quantum-safe/oqs-provider).
Things that would still need work:
- collecting/commiting to oids, especially for the hybrid algorithms
- extending the hybridization logic to RSA+PQ hybrids
- checking interoperability with other libraries implementing hybridization (eg. oqs-provider)
- change serialization formats to possible (unknown to me) "under standardization" methods


harrison314:
In this case, I would wait until the standards come out, because I have already seen three different ways to approach hybrid keys and certificates.

Issue #550, opened 2024-07-09
Consider referencing System.Memory


Rob-Hague:
Perhaps you've already considered it, but in case you haven't:

The library has many duplicate code paths, one using a Span implementation for higher targets (.NET) and one using arrays for lower targets (.NET Framework and .NET Standard 2.0)

Microsoft provide (and support) the [System.Memory](https://www.nuget.org/packages/System.Memory) package, which allows use of Span etc. on .NET Framework and .NET Standard 2.0. It is fairly common for multi-targeting libraries to reference the package on lower targets. The source code is at https://github.com/dotnet/maintenance-packages/tree/main/src/System.Memory.

This reference would allow many code paths (and public API) to be unified.

Practically, the initial change would look like:

```diff
diff --git a/crypto/src/BouncyCastle.Crypto.csproj b/crypto/src/BouncyCastle.Crypto.csproj
index 8b7c8852..166dc1ae 100644
--- a/crypto/src/BouncyCastle.Crypto.csproj
+++ b/crypto/src/BouncyCastle.Crypto.csproj
@@ -88,6 +88,10 @@
     <None Include="..\..\README.md" Pack="true" PackagePath="\" />
   </ItemGroup>

+  <ItemGroup Condition=" '$(TargetFramework)' == 'net461' or '$(TargetFramework)' == 'netstandard2.0' ">
+    <PackageReference Include="System.Memory" Version="4.5.5" />
+  </ItemGroup>
+
   <ItemGroup>
     <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3">
       <PrivateAssets>all</PrivateAssets>
```

Example unified diff (note Index/Range operators `[..32]` not supported):

```diff
diff --git a/crypto/src/crypto/modes/ChaCha20Poly1305.cs b/crypto/src/crypto/modes/ChaCha20Poly1305.cs
index 56bef5e9..ca982f9d 100644
--- a/crypto/src/crypto/modes/ChaCha20Poly1305.cs
+++ b/crypto/src/crypto/modes/ChaCha20Poly1305.cs
@@ -783,29 +783,16 @@ private ulong IncrementCount(ulong count, uint increment, ulong limit)
 
         private void InitMac()
         {
-#if NETCOREAPP2_1_OR_GREATER || NETSTANDARD2_1_OR_GREATER
             Span<byte> firstBlock = stackalloc byte[64];
             try
             {
[...]

PR #569, opened 2024-10-06
Reference System.Memory and use it in math/raw

Rob-Hague:
## Describe your changes

This change represents an initial/example contribution towards #550. It adds a reference to [System.Memory](https://www.nuget.org/packages/System.Memory) and uses it to deduplicate a number of internal code paths.

The change is kept isolated to one directory (src/math/raw) for sympathy with review. If this change is accepted, there is more that can be done in other layers of the library.

This also adds a reference to [System.ValueTuple](https://www.nuget.org/packages/System.ValueTuple) on net461 and a private reference to [PolySharp](https://github.com/Sergio0694/PolySharp). These are not necessary but allow using Index/Range syntax on lower targets.

## How has this been tested?

Existing unit tests

### DIFF
--- a/crypto/src/BouncyCastle.Crypto.csproj
+++ b/crypto/src/BouncyCastle.Crypto.csproj
@@ -6,6 +6,7 @@
     <AssemblyOriginatorKeyFile>..\..\BouncyCastle.NET.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
 	<NoWarn>1591</NoWarn>
+    <LangVersion>10.0</LangVersion>
 
     <AssemblyName>BouncyCastle.Cryptography</AssemblyName>
     <AssemblyTitle>BouncyCastle.NET Cryptography ($(TargetFramework))</AssemblyTitle>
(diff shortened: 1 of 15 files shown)


crypto/src/BouncyCastle.Crypto.csproj
```diff
@@ -6,6 +6,7 @@
     <AssemblyOriginatorKeyFile>..\..\BouncyCastle.NET.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
 	<NoWarn>1591</NoWarn>
+    <LangVersion>10.0</LangVersion>
```
Rob-Hague, line 9: Only needed for Index/Range syntax

crypto/src/BouncyCastle.Crypto.csproj
```diff
@@ -87,6 +88,13 @@
     <None Include="..\..\packageIcon.png" Pack="true" PackagePath="\" />
     <None Include="..\..\README.md" Pack="true" PackagePath="\" />
   </ItemGroup>
+    
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net461' or '$(TargetFramework)' == 'netstandard2.0' ">
+    <PackageReference Include="System.Memory" Version="4.5.5" />
+  </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net461' ">
+    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
+  </ItemGroup>
```
Rob-Hague, line 97: Only needed for Index/Range syntax

crypto/src/BouncyCastle.Crypto.csproj
```diff
@@ -101,6 +109,10 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="PolySharp" Version="1.14.1">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
```
Rob-Hague, line 115: Only needed for Index/Range syntax

uniblab:
I like updates which have more BALEETED than add.  The code we don't have is code we don't need to maintain 💯

PR #570, opened 2024-10-07
[Argon2] Multithreading support

zer0x64:
## Describe your changes

Added multithreading support for Argon2. Followup of https://github.com/bcgit/bc-csharp/pull/559

## How has this been tested?

Using the tests that were already available for Argon2.


zer0x64:
@peterdettman Here's the first follow-up PR for Argon2! The second will be for `Span` usage to avoid copies on little-endian machines

peterdettman:
Merged, thanks. I made a few followup changes, including adding a TaskFactory constructor so that parallel execution is opt-in.

PR #571, opened 2024-10-07
[Argon2] Use marshalling to avoid copies on little endian platforms

zer0x64:
## Describe your changes

Use Span and marshalling when available to avoid copies when dealing with bytes[] and ulong[] arrays in Argon2. Followup of https://github.com/bcgit/bc-csharp/pull/559

## How has this been tested?

Using the already present Argon2 test suites. I've tested it on net461 and net6.0


zer0x64:
@peterdettman Here's the second part of my Argon2 optimisations!

peterdettman:
Merged, thanks for the PR.

PR #595, opened 2025-01-27
Reintroduce and fix test for Pkcs12StoreBuilder

bb-froggy:
## Describe your changes

I found the methods DoTestSupportedTypes and basicStoreTest in Pkcs12StoreTest, which performed some IMHO useful tests on Pkcs12Store, but weren't actually part of any test (and therefore never called).

I added some code to actually call them. For this, I refactored some code into the new method CreateTestCertificateAndKey from another test to accomplish the goal with the minimum of changes possible. The other tests do not behave different now, except for one constructor call which explicitly passes a parameter that was previously not passed and the default.

I had to explicitly add the Pkcs9LocalKeyId to the X509CertificateEntry, otherwise the test failed, because the PrivateKeyEntry will always have a Pkcs9LocalKeyId added by the [Pkcs12Store](https://github.com/bcgit/bc-csharp/blob/d5d39912df2a1e8d9e9b22aecc9470210c23eff2/crypto/src/pkcs/Pkcs12Store.cs#L762-L772). I found this confusing, but maybe I am missing some detail. I consider this issue outside the scope of this change, though.

Sorry for the whitespace/tab changes that might make this change more difficult to read... the code file had a mix of tab and space indentations to begin with, so there was no consistent way to make these modifications. I could make the file use a single type of indentation in a separate patch, if you want, to make this easier to read. I only need to know whether you prefer the tabs or how many spaces?

## How has this been tested?

I ran the modified tests and they pass.

### DIFF
--- a/crypto/test/src/pkcs/test/PKCS12StoreTest.cs
+++ b/crypto/test/src/pkcs/test/PKCS12StoreTest.cs
@@ -896,7 +896,8 @@ namespace Org.BouncyCastle.Pkcs.Tests
 			AsymmetricKeyParameter	pubKey,
 			AsymmetricKeyParameter	privKey,
 			string					issuerEmail,
-			string					subjectEmail)
+			string					subjectEmail,
+			byte[]					localKeyId)
 		{
 			//
 			// distinguished name table.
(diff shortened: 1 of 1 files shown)


peterdettman:
Merged as part of #596 .

Issue #234, opened 2020-03-28
[Question] Creating a PKCS #12 File using AES


arvanauk:
I am trying to create a PKCS #12 file using AES 128.  When I create a PKCS #12 file from OpenSSL using AES and print out the structure, I get:
PKCS7 PBES2, PBKDF2, AES-128-CDC, Iteration 4096, PRF HMACwithSHA1.

I get errors when I try to set the key and certificate algorithm in the PKCS12StoreBuilder.  There are no examples using the PKCS12Store and PKCS12StoreBuilder classes.

I am using version 1.8.6.

jonproch117:
I was able to set the key algorithm on the PKCS12StoreBuilder using `.SetKeyAlgorithm(NistObjectIdentifiers.IdAes128Cbc, PkcsObjectIdentifiers.IdHmacWithSha1)`.

The certificate algorithm does not appear to currently support PBES2-based encryption schemes, just the legacy PKCS12 ones.

PR #596, opened 2025-01-28
Add AES support for PKCS12StoreBuilder

bb-froggy:
## Describe your changes

The PKCS12StoreBuilder lacks support and documented examples for creating PKCS#12 with modern algorithms, e.g. AES. This is described in #234 .

Specifically, the key encryption with AES (e.g. PBES2 with AES + PBKDF2 with SHA) works, but has no test, while the certificate encryption seems to require legacy algorithms and seemingly doesn't support using no encryption. The latest OpenSSL versions prohibit the use of legacy cryptography, so I believe they cannot open these PKCS#12 files.

This change adds a test for creating a PKCS#12 with AES. This is in the first commit already.

I am working on getting rid of the legacy cryptography for the certificate encryption, but it is not in the first commit of this WIP PR. I would add support to not encrypt the certificates (which is often okay IMHO) or allow using modern algorithms.

The code change founds on another PR, #595 to allow testing the new code.

## How has this been tested?

There is a new test for AES. It currently succeeds, and I should succeed without legacy crypto for the certificate encryption before merging.

- [ ] I have performed a self-review of my code
- [ ] I have kept the patch limited to only change the parts related to the patch

### DIFF
--- a/crypto/test/src/pkcs/test/PKCS12StoreTest.cs
+++ b/crypto/test/src/pkcs/test/PKCS12StoreTest.cs
@@ -896,7 +896,8 @@ namespace Org.BouncyCastle.Pkcs.Tests
 			AsymmetricKeyParameter	pubKey,
 			AsymmetricKeyParameter	privKey,
 			string					issuerEmail,
-			string					subjectEmail)
+			string					subjectEmail,
+			byte[]					localKeyId)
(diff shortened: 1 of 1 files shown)


peterdettman:
Merged, thanks for the PR.

PR #612, opened 2025-05-05
Update MemoryMarshal.Write calls

uniblab:
## Describe your changes

Update MemoryMarshal.Write calls to declare the 'value' parameter as IN not REF in net8+ compilations.  MS changed the semantics of the function in net8+ releases.

## How has this been tested?

All tests of the included test project passed w/o error.


peterdettman:
Seems like reasonable preparation for when we add net80 target. Merged, thanks.

PR #628, opened 2025-06-11
Handle `keytool` v20+ change to `AlgorithmIdentifier`

spbsoluble:
## Describe your changes

This is to handle a change in ASN.1 encoding introduced in keytool v20 where the `AlgorithmIdentifier` omits the NULL parameters(/DerNull.Instance). 

keytool <=v19: `[1.3.6.1.4.1.42.2.17.1.1, NULL]  
keytool >=v20 `[1.3.6.1.4.1.42.2.17.1.1]`

## How has this been tested?

[gen_keys.txt](https://github.com/user-attachments/files/20699080/gen_keys.txt)
Generated key types RSA, EC and ED using keytool v19-v23 and was able to decode key+chain into PEM resolving the "unknown algorithm" exception previously encountered.


spbsoluble:
@peterdettman I totally missed this because I didn't test with < .netstandard2.1. I've updated the PR accordingly, thanks for calling that out

PR #645, opened 2025-08-06
Expose the S2K iteration count when symmetrically encrypting with OpenPGP

JonathanMatthewsEnable:
## Describe your changes

The OpenPGP message format allows the number of S2K iterations to be specified when creating a symmetrically encrypted message. However, the ability to set this was not previously exposed via the BouncyCastle.Cryptography C# library and a default coded count of 0x60 (decimal 96) was always used, corresponding to 65536 cycles.

This change modifies the `AddMethod`, `AddMethodUtf8`, and `AddMethodRaw` methods of the `PgpEncryptedDataGenerator` class to allow passing in a coded S2K iteration count.

## How has this been tested?

A local build of the nuget package was performed and consumed by a dependent application, which successfully made use of the now exposed S2K cycle parameter when creating a symmetrically encrypted PGP message. The cycle count passed in was confirmed to be respected using `gpg --list-packets` on the encrypted message.

All non-explicit automated tests are passing for the following targets: net461, net472, net6.0, netcoreapp3.1

### DIFF
--- a/crypto/src/openpgp/PgpEncryptedDataGenerator.cs
+++ b/crypto/src/openpgp/PgpEncryptedDataGenerator.cs
@@ -329,7 +329,37 @@ namespace Org.BouncyCastle.Bcpg.OpenPgp
         /// Conversion of the passphrase characters to bytes is performed using Convert.ToByte(), which is
         /// the historical behaviour of the library (1.7 and earlier).
         /// </remarks>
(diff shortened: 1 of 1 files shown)


peterdettman:
Adding the extra parameter in this way would be a breaking change. Please change the patch to add three new methods instead.

peterdettman:
Merged, thanks for the PR.

PR #648, opened 2025-09-15
Fixes Zlib compression issues.

GitHubProUser67:
The current zlib implementation is "mostly" fine, but some edge cases are not simulated properly.

This commit effectively puts it to parity with the C++ version, and allows this implementation to be used to open and compress "EdgeZlib" files used by various PS3 games.

## How has this been tested?

Compression/decompression of EdgeZlib files, Quazal packets and deflate http data.

Check this repo (where I used this change to remove a whoping 3 Zlib duplicata libraries (because none was right...)): https://github.com/GitHubProUser67/MultiServer3/tree/Super-Branch/BackendServices/CastleLibrary

- [ x] I have performed a self-review of my code
- [ x] I have kept the patch limited to only change the parts related to the patch

### DIFF
--- a/crypto/src/util/zlib/ZInputStream.cs
+++ b/crypto/src/util/zlib/ZInputStream.cs
@@ -51,7 +51,7 @@ namespace Org.BouncyCastle.Utilities.Zlib
             return z;
         }
 
-        private const int BufferSize = 4096;
+        private const int BufferSize = 512;
 
         protected ZStream z;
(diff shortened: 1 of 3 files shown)


peterdettman:
Thanks for the PR. I don't think we want the ZStream change (because of Console output - is an exception warranted?), and is the ZInputStream change to the buffer size really needed?

Is it possible to provide example input/output cases that fail without this PR?

PR #670, opened 2026-04-01
Enhance PKCS#8 support: add .NET X509Certificate2 interop and fix RSA…

KonradSop:
### Description
This PR addresses several issues regarding PKCS#8 and X.509 interoperability brought up in the community discussions, specifically regarding RSA key encoding and usage convenience.

### Related Discussion
- Fixes/Addresses issues raised in Discussion #592

### Changes
1. **New DotNetUtilities Shortcuts**: Added `GetSubjectPublicKeyInfo` and `GetSubjectPublicKeyInfoDer` methods to simplify extracting PKCS#8/X.509 public key information directly from .NET `X509Certificate2` objects in a single line.
2. **Standardized RSA Encoding**: Verified and documented that `SubjectPublicKeyInfoFactory` correctly includes mandatory `DER NULL (05 00)` parameters for RSA AlgorithmIdentifiers (OID 1.2.840.113549.1.1.1), resolving the "missing parameter" issue some users encountered when interoping with other crypto providers.
3. **Professional Documentation**: 
   - Converted legacy Java-style comments into standard C# triple-slash XML documentation across `DotNetUtilities`, `SubjectPublicKeyInfoFactory`, and `PrivateKeyInfoFactory`.
   - Added usage `<example>` snippets for common tasks.
4. **Unit Tests**: Added `Pkcs8Test.cs` to verify correct ASN.1 encoding (specifically RSA NULL parameters) and test the new utility methods across all supported .NET frameworks.

### Checklist before requesting a review
- [x] I have kept the patch limited to only change the parts related to the patch (Ensured no .gitignore slip-ups this time!)
- [x] This change requires a documentation update (n/a - this PR itself IS a documentation update)


crypto/src/security/DotNetUtilities.cs
```diff
@@ -54,17 +54,71 @@ public static SystemX509.X509Certificate ToX509Certificate(X509CertificateStruct
...
+        /// and then extracts the public key information using <see cref="SubjectPublicKeyInfoFactory"/>.
+        /// </remarks>
+        /// <exception cref="ArgumentNullException">If <paramref name="certificate"/> is null.</exception>
+        public static SubjectPublicKeyInfo GetSubjectPublicKeyInfo(SystemX509.X509Certificate2 certificate)
+        {
+            if (certificate == null)
+                throw new ArgumentNullException(nameof(certificate));
+
+            var bcCert = FromX509Certificate(certificate);
+            return SubjectPublicKeyInfoFactory.CreateSubjectPublicKeyInfo(bcCert.GetPublicKey());
```
peterdettman, line 89: From .NET 6.0 could use:
```
return SubjectPublicKeyInfo.GetInstance(certificate.PublicKey.ExportSubjectPublicKeyInfo());
```

crypto/src/security/DotNetUtilities.cs
```diff
@@ -54,17 +54,71 @@ public static SystemX509.X509Certificate ToX509Certificate(X509CertificateStruct
...
+        /// <param name="certificate">The .NET certificate.</param>
+        /// <returns>A byte array containing the DER-encoded public key info.</returns>
+        /// <remarks>
+        /// This is a convenience method that returns the raw DER-encoded bytes of the public key info,
+        /// suitable for saving to disk or transmitting over the network.
+        /// </remarks>
+        /// <exception cref="ArgumentNullException">If <paramref name="certificate"/> is null.</exception>
+        public static byte[] GetSubjectPublicKeyInfoDer(SystemX509.X509Certificate2 certificate)
+        {
+            return GetSubjectPublicKeyInfo(certificate).GetEncoded(Asn1Encodable.Der);
```
peterdettman, line 108: From .NET 6.0 could use:
```
return certificate.PublicKey.ExportSubjectPublicKeyInfo();
```

crypto/src/security/DotNetUtilities.cs
```diff
@@ -54,17 +54,71 @@ public static SystemX509.X509Certificate ToX509Certificate(X509CertificateStruct
...
+        /// <summary>
+        /// Create a Bouncy Castle <see cref="X509Certificate"/> from a .NET <see cref="SystemX509.X509Certificate2"/>.
+        /// </summary>
+        /// <param name="x509Cert">The .NET certificate.</param>
+        /// <returns>A Bouncy Castle <see cref="X509Certificate"/>.</returns>
         public static X509Certificate FromX509Certificate(SystemX509.X509Certificate2 x509Cert) =>
             new X509Certificate(x509Cert.RawData);
 
+        /// <summary>
+        /// Extract the <see cref="SubjectPublicKeyInfo"/> (X.509 / PKCS#8) from a .NET <see cref="SystemX509.X509Certificate2"/>.
```
peterdettman: "PKCS#8" doesn't really belong here. SubjectPublicKeyInfo is an ASN.1 type from the X.509 standard and is used for public keys. PKCS#8 is a separate standard for private keys.

crypto/src/x509/SubjectPublicKeyInfoFactory.cs
```diff
@@ -17,8 +17,12 @@
 namespace Org.BouncyCastle.X509
 {
     /// <summary>
-    /// A factory to produce Public Key Info Objects.
+    /// A factory to produce SubjectPublicKeyInfo (X.509 / PKCS#8) objects from Bouncy Castle public key parameters.
```
peterdettman: ~PKCS#8~ again.

crypto/src/x509/SubjectPublicKeyInfoFactory.cs
```diff
@@ -31,11 +35,19 @@ public static class SubjectPublicKeyInfoFactory
...
 
         /// <summary>
-        /// Create a Subject Public Key Info object for a given public key.
+        /// Create a <see cref="SubjectPublicKeyInfo"/> object for a given public key.
         /// </summary>
-        /// <param name="publicKey">One of ElGammalPublicKeyParameters, DSAPublicKeyParameter, DHPublicKeyParameters, RsaKeyParameters or ECPublicKeyParameters</param>
-        /// <returns>A subject public key info object.</returns>
-        /// <exception cref="Exception">Throw exception if object provided is not one of the above.</exception>
+        /// <example>
+        /// Example of converting a .NET X509Certificate2 to a PKCS#8/DER byte array:
```
peterdettman: ~PKCS#8~ again.

crypto/test/src/pkcs/test/Pkcs8Test.cs
```diff
@@ -0,0 +1,101 @@
...
+using Org.BouncyCastle.X509;
+
+// Use aliases to avoid ambiguity between BC and .NET X509 types
+using BcX509Certificate = Org.BouncyCastle.X509.X509Certificate;
+using SystemX509 = System.Security.Cryptography.X509Certificates;
+
+namespace Org.BouncyCastle.Pkcs.Tests
+{
+    [TestFixture]
+    public class Pkcs8Test
```
peterdettman: This test name (because of the Pkcs8) and location makes little sense.

I suggest adding `TestDotNetUtilitiesGetSubjectPublicKeyInfoDer` to `Org.BouncyCastle.Security.Tests.TestDotNetUtilities` instead (include the whole test case under the `#if`). The other two could perhaps just be added to `Org.BouncyCastle.Security.TestsTestEncodings` or, if you prefer,  a new test class in `Org.BouncyCastle.X509.Tests`.

peterdettman (changes requested):

KonradSop:
Thanks for the detailed review! All six points addressed in the latest push:

1. GetSubjectPublicKeyInfo — now uses SubjectPublicKeyInfo.GetInstance(certificate.PublicKey.ExportSubjectPublicKeyInfo()) under #if NET6_0_OR_GREATER, with the BC round-trip kept as fallback for netstandard2.0 and net461. Confirmed via the MS Learn "Applies to" section that PublicKey.ExportSubjectPublicKeyInfo() was introduced in .NET 6.0, matching your suggestion exactly.
2. GetSubjectPublicKeyInfoDer — now returns certificate.PublicKey.ExportSubjectPublicKeyInfo() directly under the same guard.
3. DotNetUtilities.GetSubjectPublicKeyInfo XML summary — "PKCS#8" removed; now describes SubjectPublicKeyInfo as an X.509 ASN.1 type for public keys.
4. SubjectPublicKeyInfoFactory class summary — "PKCS#8" removed.
5. <example> caption in CreateSubjectPublicKeyInfo — "PKCS#8" removed; now says "DER-encoded SubjectPublicKeyInfo byte array".
6. Tests — Pkcs8Test.cs deleted. TestGetSubjectPublicKeyInfoDer moved into TestDotNetUtilities (wrapped in #if NET6_0_OR_GREATER). The other two tests moved into TestEncodings (as TestRsaPublicKeyInfoEncodingHasNullParameters and TestSubjectPublicKeyInfoFactoryRsaConsistency).

Small deviation worth flagging: I dropped the DotNetUtilities prefix from the moved test name since the enclosing class is already TestDotNetUtilities — existing tests in that class follow the same convention (TestECDsaInterop, TestRsaInterop, etc.). Happy to rename back if you prefer.

peterdettman:
Thanks for the update. Squashed, rebased, merged.

PR #708, opened 2026-09-01
fix: Regression: Use Modulus and Exponent for RSA Hash Codes

bb-froggy:
Before BC 2.7.0, GetHashCode of RsaKeyParameters XORed the modulus and the exponent of the key as well as whether it was private.

With 8ee52127e61fb4e61d97e868ba74323d3b581c57, this changed to XORing the modulus with itself (always resulting in 0) and so the value of GetHashCode() just depends on whether the key is private, but not on the actual parameters.

## Describe your changes

This fix reverts to the previous behavior, where GetHashCode XORs modulus, exponent, and whether the key is private.

## How has this been tested?

I haven't actually tested it.


peterdettman:
Merged, thanks for the PR.

